PR #4, opened 2022-11-14
feat(core): better caret position management + new class `MaskModel`

nsbarsukov:


splincode (approved):

projects/core/src/lib/classes/mask-model/mask-model.ts
```diff
@@ -0,0 +1,88 @@
...
+        const [unmaskedFrom, unmaskedTo] = unmaskedSelection;
+        const isOnlyFixedCharsSelected = unmaskedFrom === unmaskedTo;
+        const isLastChar = unmaskedTo >= unmaskedValue.length;
+
+        if (isOnlyFixedCharsSelected && !isLastChar) {
+            this.caretIndex = selectedRange[0];
+            return;
+        }
+
+        const newUnmaskedValue =
```
waterplea, line 74: This big ternary is hard to read, let's move it to a separate constant:
```ts
const newFrom = isOnlyFixedCharsSelected && isLastChar ? unmaskedFrom - 1 : unmaskedFrom;
```
Hope it fits in one line :D maybe `isOnlyFixedCharsSelected` can drop `is` or made shorter some other way to keep such things in one line — it's easier to follow this way :)

projects/core/src/lib/mask.ts
```diff
@@ -50,34 +53,72 @@ export class Mask {
...
 
-    private fillInputWithFixedValues(isBackspace: boolean = false): void {
-        const {value} = this.elementRef;
-        const modifiedValue = adjustFixedMaskCharacters(
-            value,
-            this.maskExpression,
-            isBackspace,
-        );
+    private updateValue(newValue: string): void {
+        const {elementRef} = this;
```
waterplea, line 110: Not sure this saves any chars :D

projects/core/src/lib/mask.ts
```diff
@@ -50,34 +53,72 @@ export class Mask {
...
+
+        if (elementRef.value !== newValue) {
+            elementRef.value = newValue;
+        }
+    }
+
+    private updateCaretIndex(newCaretIndex: number): void {
+        const {elementRef} = this;
```
waterplea, line 118: And here :) I have a feeling that less lines would be easier to scan and follow.

PR #10, opened 2022-11-16
feat(core): handle `Paste`-event

nsbarsukov:
## What is the new behavior?
1. Handle `Paste`-event
2. Rename public API: `Mask` => `Maskito` & `MaskOptions` => `MaskitoOptions`
3. Do not prevent events if it is not required


splincode (approved):

vladimirpotekhin (approved):

projects/core/src/lib/mask.ts
```diff
@@ -53,72 +62,89 @@ export class Mask {
...
-        event.preventDefault();
-
-        const {selectionStart, selectionEnd, value: initialValue} = this.elementRef;
-        const maskModel = new MaskModel(
-            initialValue,
-            selectionStart ?? 0,
-            maskExpression,
-        );
+        const maskModel = new MaskModel(element.value, selectionStart, maskExpression);
+        const [from, to]: [number, number] =
```
waterplea, line 92: Is type here necessary?

projects/core/src/lib/mask.ts
```diff
@@ -53,72 +62,89 @@ export class Mask {
...
+        }
     }
 
     private handlePaste(event: ClipboardEvent): void {
-        // TODO: finish later
-        // eslint-disable-next-line no-console
-        console.log('===[handlePaste]===', event.cancelable, getClipboardDataText(event));
+        const {element, maskExpression, selectionStart, selectionEnd} = this;
+        const maskModel = new MaskModel(element.value, selectionStart, maskExpression);
+        const insertedText = event.clipboardData?.getData('text/plain') ?? '';
```
waterplea, line 113: Let's use `||` for smaller transpiled bundle

PR #15, opened 2022-11-18
refactor(core): introduce concept `ElementState` + replace `caret` with `selection`

nsbarsukov:
```ts
export interface ElementState {
    value: string;
    selection: [from: number, to: number];
}

```

### DIFF
--- a/projects/core/src/lib/classes/mask-model/mask-model.ts
+++ b/projects/core/src/lib/classes/mask-model/mask-model.ts
@@ -1,54 +1,58 @@
-import {MaskExpression} from '../../types';
+import {ElementState, MaskExpression, MaskitoOptions} from '../../types';
 import {validateValueWithMask} from './utils/validate-value-with-mask';
 import {removeFixedMaskCharacters} from './utils/remove-fixed-mask-characters';
 import {addFixedMaskCharacters} from './utils/add-fixed-mask-characters';
 
 export class MaskModel {
     value = '';
+    // TODO: refactor code and replace `caretIndex: number` with `selection: [from: number, to: number]`
     caretIndex = 0;
 
-    constructor(
-        readonly initialValue: string,
-        readonly initialCaretIndex: number,
-        private readonly mask: MaskExpression,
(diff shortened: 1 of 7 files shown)


projects/core/src/lib/types/element-state.ts
```diff
@@ -0,0 +1,4 @@
+export interface ElementState {
```
splincode, line 1: MaskitoElementState

splincode (approved):

projects/core/src/lib/types/element-state.ts
nsbarsukov: This interface is not Public API. Should we rename it ?

nsbarsukov: Public API:
https://github.com/Tinkoff/maskito/blob/80384b30204eeacde1c7318efa36faf5cf477c5e/projects/core/src/index.ts#L1-L2

splincode: as you wish))

nsbarsukov: I think the shorter name is better)

The prefix `Maskito` don't add anything to the meaning of naming when you read code.

But public API should be all prefixed, of course!

projects/core/src/lib/classes/mask-model/mask-model.ts
```diff
@@ -1,54 +1,58 @@
...
-        newCharacters: string,
-    ): void {
-        const {unmaskedValue, unmaskedSelection} = removeFixedMaskCharacters(
-            this.value,
-            this.mask,
-            selectedRange,
+    addCharacters(selection: [from: number, to: number], newCharacters: string): void {
+        const unmaskedElementState = removeFixedMaskCharacters(
+            {value: this.value, selection},
+            this.maskExpression,
```
waterplea, line 33: You invoke this getter multiple times and it might be expensive, depending on the mask function. Let's save it to a const.

PR #17, opened 2022-11-18
fix(core): incorrect caret position (case with range selection + new key)

nsbarsukov:
https://github.com/Tinkoff/maskito/blob/d03134c160f8a4e6d1cc85d6130b5ca8f0924410/projects/demo-integrations/cypress/tests/sandbox/input-phone.cy.ts#L279-L307

### DIFF
--- a/projects/core/src/lib/classes/mask-model/utils/add-fixed-mask-characters.ts
+++ b/projects/core/src/lib/classes/mask-model/utils/add-fixed-mask-characters.ts
@@ -16,34 +16,28 @@ export function addFixedMaskCharacters(
     const maskedValue = [
         ...Array.from(initialElementState.value),
         '', // extra iteration to take all tailed fixed characters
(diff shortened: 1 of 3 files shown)


splincode (approved):

projects/core/src/lib/classes/mask-model/utils/add-fixed-mask-characters.ts
```diff
@@ -9,47 +9,68 @@ export function addFixedMaskCharacters(
...
+    mask: Array<RegExp | string>,
+    formattedValue: string,
+    typedCharacter: string,
+): string {
+    let leadingFixedCharacters = ``;
+
+    for (let i = formattedValue.length; i < mask.length; i++) {
+        const charConstraint = mask[i];
+
+        if (!isFixedCharacter(charConstraint)) {
```
waterplea, line 64: These two ifs can be combined into one :)

PR #28, opened 2022-12-02
feat(core): introduce concepts `preprocessor` / `postprocessor`

nsbarsukov:
## New API description
This PR introduces new concepts: **preprocessor** and **postprocessor**.
Both utility functions become a part of `MaskitoOptions`:
```ts
export interface MaskitoOptions {
    mask: MaskExpression | (() => MaskExpression);
    preprocessor?: MaskPreprocessor;
    postprocessor?: MaskPostprocessor;
}
```

## Preprocessor
This function is executed every time BEFORE mask validation starts.
You can modify any new value before it will be rejected after mask validation.

```ts
const ONLY_DIGITS_REG_EXP = /^\d+(\.\d*)?$/;

const maskedInput = new Maskito(element, {
    mask: ONLY_DIGITS_REG_EXP,
    preprocessor: ({elementState, data}) => {
        const {value, selection} = elementState;

        return {
            elementState: {
                selection,
                value: value.replaceAll(',', '.'),
            },
            data: data?.replaceAll(',', '.'),
        };
    },
})
```

## Postprocessor
This function is executed every time AFTER mask validation ends.
The value (returned from your postprocessor) will not be validated by mask.
This function gives you maximum flexibility for all your edge-cases.

```ts
const PHONE_MASK = [
    '+', '7', '(', /\d/, /\d/, /\d/, ')', /\d/, /\d/, /\d/, '-', /\d/, /\d/, '-', /\d/, /\d/
];
const countryPrefix = '+7 ';
const minCaretIndex = countryPrefix.length;

const maskedInput = new Maskito(element, {
    mask: PHONE_MASK,
    postprocessor: ({value, selection}) => {
        return value.length > minCaretIndex
            ? {selection, value}
            : {selection: [minCaretIndex, minCaretIndex], value: countryPrefix};
        },
    };
)
```


vladimirpotekhin (approved):

projects/core/src/lib/utils/element-states-equality.ts
```diff
@@ -0,0 +1,8 @@
+import {ElementState} from '../types';
+
+export function areElementValuesEqual(
+    sampleState: ElementState,
+    ...states: ElementState[]
+): boolean {
+    return states.every(state => state.value === sampleState.value);
```
waterplea, line 7: Can destruct here

projects/core/src/lib/types/mask-processors.ts
```diff
@@ -0,0 +1,8 @@
+import {ElementState} from './element-state';
+
+export type MaskPreprocessor = (_: {elementState: ElementState; data?: string}) => {
+    elementState: ElementState;
+    data?: string;
```
waterplea, line 5: I think it would be easier for people not to test it for `undefined` all the time, we can just pass empty string for those cases.

projects/core/src/lib/types/mask-processors.ts
```diff
@@ -0,0 +1,8 @@
...
+};
+
+export type MaskPostprocessor = (elementState: ElementState) => ElementState;
```
waterplea, line 8: This might need `rawState` as well, to know what state was before mask did its job.

projects/core/src/lib/utils/element-states-equality.ts
nsbarsukov: https://github.com/Tinkoff/maskito/pull/47

projects/core/src/lib/types/mask-processors.ts
nsbarsukov: https://github.com/Tinkoff/maskito/pull/47

PR #37, opened 2022-12-08
feat: new `Time`-mask (`kit`) + new `overwriteMode`-mode (`core`)

### LINKED ISSUE #30: 🚀 - `Time`-mask (`@maskito/kit`)
Developing mask for [InputTime](https://taiga-ui.dev/components/input-time)-component.

## Proposed API
```ts
import {Maskito} from '@maskito/core';
import {maskitoTimeOptionsGenerator} from '@maskito/kit';

const inputElement = document.querySelector('input#time');

const inputTime = new Maskito(
    inputElement,
    maskitoTimeOptionsGenerator({
        mode: 'HH:MM:SS',
        maxValues: {HH: 23, MM: 59, SS: 59, MS: 999}, // default value
        // ... (more options?)
    }),
);
```

**Problems:**
* It is not obvious for user what `maxValues`-option means!
For example, `maxValues: {HH: 11, MM: 30}`. Okey, `11:31` is obviously invalid value. What's about `10:31` ?
We should think about better naming (now it can be confused with max allowable point in the time).

## Acceptance criteria
- [ ] Minutes and seconds cannot be greater than or `59`. Millisecond – `999`.
- [ ] Hour cannot be greater than `maxValues['HH']`, minutes – `maxValues['MM']`, seconds – `maxValues['SS']`, millisecond - `maxValues['MS']`.
- [ ] If the first typed digit for time-segment (`hours` |  `minutes` | `seconds` | `ms`) is already exceeds limits, mask adds `0`.
For example, you type `11:7`. Seventy-hours are not allowable. Mask should guess that you mean `11:07`.
- [ ] User cannot type anything excepts digits (and fixed values, of course).

nsbarsukov:
## PR Checklist

- [X] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #30

### DIFF
--- a/projects/core/src/lib/mask.ts
+++ b/projects/core/src/lib/mask.ts
@@ -121,6 +121,7 @@ export class Maskito extends MaskHistory {
     private conformValueToMask(): void {
         const {elementState} = this.options.preprocessor({
             elementState: this.elementState,
+            data: '',
         });
         const maskModel = new MaskModel(elementState, this.options);
         const {value, selection} = this.options.postprocessor(maskModel);
@@ -132,6 +133,7 @@ export class Maskito extends MaskHistory {
     private handleDelete(event: Event, isForward: boolean): void {
         const {elementState} = this.options.preprocessor({
             elementState: this.elementState,
+            data: '',
         });
         const [from, to] = extendToNotEmptyRange(elementState.selection, isForward);
         const maskModel = new MaskModel(elementState, this.options);
@@ -173,9 +175,7 @@ export class Maskito extends MaskHistory {
 
         const [from, to] = elementState.selection;
         const newPossibleValue =
-            elementState.value.slice(0, from) +
-            insertedText +
-            elementState.value.slice(to);
+            elementState.value.slice(0, from) + data + elementState.value.slice(to);
         const {value, selection} = this.options.postprocessor(maskModel);
 
         if (newPossibleValue !== value) {
--- a/projects/core/src/lib/mask.ts
+++ b/projects/core/src/lib/mask.ts
@@ -122,6 +122,7 @@ export class Maskito extends MaskHistory {
         const {elementState} = this.options.preprocessor({
             elementState: this.elementState,
             data: '',
+            actionType: 'validation',
         });
         const maskModel = new MaskModel(elementState, this.options);
         const {value, selection} = this.options.postprocessor(maskModel);
@@ -131,18 +132,24 @@ export class Maskito extends MaskHistory {
     }
 
     private handleDelete(event: Event, isForward: boolean): void {
+        const initialState: ElementState = {
+            value: this.elementState.value,
+            selection: extendToNotEmptyRange(this.elementState.selection, isForward),
+        };
         const {elementState} = this.options.preprocessor({
-            elementState: this.elementState,
+            elementState: initialState,
             data: '',
+            actionType: isForward ? 'deleteForward' : 'deleteBackward',
         });
-        const [from, to] = extendToNotEmptyRange(elementState.selection, isForward);
         const maskModel = new MaskModel(elementState, this.options);
+        const [from, to] = elementState.selection;
 
         maskModel.deleteCharacters([from, to]);
 
         const newElementState = this.options.postprocessor(maskModel);
         const newPossibleValue =
-            elementState.value.slice(0, from) + elementState.value.slice(to);
+            initialState.value.slice(0, initialState.selection[0]) +
+            initialState.value.slice(initialState.selection[1]);
 
         if (newPossibleValue === newElementState.value) {
             return;
@@ -150,7 +157,9 @@ export class Maskito extends MaskHistory {
 
         event.preventDefault();
 
-        if (areElementValuesEqual(elementState, maskModel, newElementState)) {
+        if (
+            areElementValuesEqual(initialState, elementState, maskModel, newElementState)
+        ) {
             // User presses Backspace/Delete for the fixed value
             return this.updateSelectionRange(isForward ? [to, to] : [from, from]);
         }
@@ -164,6 +173,7 @@ export class Maskito extends MaskHistory {
         const {elementState, data: insertedText = data} = this.options.preprocessor({
             data,
             elementState: this.elementState,
+            actionType: 'insert',
         });
         const maskModel = new MaskModel(elementState, this.options);
 
--- a/projects/core/src/lib/mask.ts
+++ b/projects/core/src/lib/mask.ts
@@ -136,6 +136,7 @@ export class Maskito extends MaskHistory {
             value: this.elementState.value,
             selection: extendToNotEmptyRange(this.elementState.selection, isForward),
         };
+        const [initialFrom, initialTo] = initialState.selection;
         const {elementState} = this.options.preprocessor({
             elementState: initialState,
             data: '',
@@ -148,8 +149,8 @@ export class Maskito extends MaskHistory {
 
         const newElementState = this.options.postprocessor(maskModel);
         const newPossibleValue =
-            initialState.value.slice(0, initialState.selection[0]) +
-            initialState.value.slice(initialState.selection[1]);
+            initialState.value.slice(0, initialFrom) +
+            initialState.value.slice(initialTo);
 
         if (newPossibleValue === newElementState.value) {
             return;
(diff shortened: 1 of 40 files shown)


splincode (approved):

projects/core/src/lib/mask.ts
```diff
@@ -130,25 +132,35 @@ export class Maskito extends MaskHistory {
...
 
         if (newPossibleValue === newElementState.value) {
             return;
         }
 
         event.preventDefault();
 
-        if (areElementValuesEqual(elementState, maskModel, newElementState)) {
+        if (
+            areElementValuesEqual(initialState, elementState, maskModel, newElementState)
```
splincode, line 162: what about
`areElementValuesEqual({ initialState, elementState, maskModel, newElementState })`

splincode (approved):

projects/kit/src/lib/time/utils/parse-time-string.ts
```diff
@@ -0,0 +1,15 @@
+import {MaskitoTimeSegments} from '../types';
+
+/**
+ * @param timeString can be with/without fixed characters
+ */
+export function parseTimeString(timeString: string): MaskitoTimeSegments {
```
waterplea, line 6: This function looks like it works with hours down to milliseconds with optional precision, but what if we want to go the other way around and store seconds + milliseconds and optionally add minutes and hours?

projects/kit/src/lib/time/types/time-segments.ts
```diff
@@ -0,0 +1,6 @@
+export interface MaskitoTimeSegments<T = string> {
```
waterplea, line 1: What's the point of generic here? Shouldn't it just be `string | number`?

projects/kit/src/lib/time/types/time-mode.ts
```diff
@@ -0,0 +1 @@
+export type MaskitoTimeMode = 'HH:MM' | 'HH:MM:SS' | 'HH:MM:SS.MSS';
```
waterplea, line 1: That's what I asked about in parsing function. I think we should also support `SS.MSS` and `MM.SS.MSS`. Maybe in the next iteration, but keep that in mind.

projects/kit/src/lib/time/constants/default-time-segment-max-values.ts
```diff
@@ -0,0 +1,8 @@
+import {MaskitoTimeSegments} from '../types';
+
+export const DEFAULT_TIME_SEGMENT_MAX_VALUES: MaskitoTimeSegments<number> = {
+    hours: 23,
```
waterplea, line 4: Just something to keep in mind, we'd want to support AM/PM mode eventually too.

waterplea:
image
Pressing '2' does not do anything, should move caret behind '2'.

waterplea:
image
Pressing 1 moves all other digits to the left, making them invalid and thus drops all remaining characters. Not sure this is the best way to handle this. Maybe replace second char with 0 so numbers don't shift?

waterplea:
image

Pressing `1` doesn't do anything.

projects/core/src/lib/mask.ts
```diff
@@ -173,9 +186,7 @@ export class Maskito extends MaskHistory {
 
         const [from, to] = elementState.selection;
         const newPossibleValue =
-            elementState.value.slice(0, from) +
-            insertedText +
-            elementState.value.slice(to);
+            elementState.value.slice(0, from) + data + elementState.value.slice(to);
```
waterplea, line 189: What is the difference between `insertedText` and `data` in this case?

projects/demo/src/pages/kit/time/time-mask-doc.component.ts
```diff
@@ -0,0 +1,42 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TimeMaskDocComponent implements GeneratorOptions {
+    readonly preparedMasks = {
+        'HH:MM': maskitoTimeOptionsGenerator({mode: 'HH:MM'}),
+        'HH:MM:SS': maskitoTimeOptionsGenerator({mode: 'HH:MM:SS'}),
+        'HH:MM:SS.MSS': maskitoTimeOptionsGenerator({mode: 'HH:MM:SS.MSS'}),
+        'HH:MM 12-hours': maskitoTimeOptionsGenerator({
+            mode: 'HH:MM',
+            timeSegmentMaxValues: {hours: 12},
```
waterplea, line 24: Maybe just `maxValues`?

projects/kit/src/lib/time/processors/zero-placeholders-preprocessor.ts
```diff
@@ -0,0 +1,21 @@
+import {MaskitoOptions} from '@maskito/core';
+
+export const zeroPlaceholdersPreprocessor: NonNullable<
```
waterplea, line 3: Just a thought: I wonder if we would need to add `pre/postprocessorPipe` utility function to pipe multiple processors in order.

projects/kit/src/lib/time/time-mask.ts
```diff
@@ -0,0 +1,46 @@
...
+        ...DEFAULT_TIME_SEGMENT_MAX_VALUES,
+        ...timeSegmentMaxValues,
+    };
+
+    return {
+        mask: Array.from(mode).map(char =>
+            TIME_FIXED_CHARACTERS.includes(char) ? char : /\d/,
+        ),
+        /**
+         * TODO: create new utility
```
waterplea, line 27: Just added a comment about this :D Removed it since it's already mentioned.

projects/kit/src/lib/time/processors/zero-padding-postprocessor.ts
```diff
@@ -0,0 +1,46 @@
...
+        const [from] = selection;
+
+        return {
+            value: newTimeString,
+            selection: [from + newCharactersAmount, from + newCharactersAmount],
+        };
+    };
+}
+
+function padWithZeroes(timeSegment: string, maxSegmentValue: string): string {
```
waterplea, line 38: Shouldn't it just be [padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) somehow?

projects/core/src/lib/mask.ts
nsbarsukov: Order of arguments doesn't matter for this function)

https://github.com/Tinkoff/maskito/blob/1855c3587b70522b0d1c196d1fe036e8bc11ae71/projects/core/src/lib/utils/element-states-equality.ts#L3-L6

projects/core/src/lib/mask.ts
nsbarsukov: * `data` - inserted characters BEFORE preprocessing (raw data from `Event.data`).
* `insertedText` - inserted characters AFTER preprocessing

We calculates `newPossibleValue` to decide if we have to prevent native event? Will natively typed character make mask valid (without programatic interference)?

It is especially useful for `RegExp`-mask (without fixed characters masks).
The majority of all typed characters can be inserted natively (without programatic value patching) ?

projects/kit/src/lib/time/processors/zero-padding-postprocessor.ts
nsbarsukov: It can be possible if all time segments are the same length (but they can be 2 or 3-character length).

That is why I need recursive function here.

projects/kit/src/lib/time/types/time-segments.ts
nsbarsukov: This type is used many times in the code.
Sometimes it is convenient to have object with all strings. Sometimes with all numbers.
And don't do extra type checks.

projects/kit/src/lib/time/types/time-mode.ts
nsbarsukov: Okey, let's keep it in mind for the next iteration.

For the first iteration, all `kit`-masks should reproduce already existing Taiga UI features + fixing old bugs.
All new features – next iteration.

projects/kit/src/lib/time/constants/default-time-segment-max-values.ts
nsbarsukov: https://github.com/Tinkoff/maskito/pull/37#discussion_r1046913776

projects/kit/src/lib/time/types/time-mode.ts
nsbarsukov: - https://github.com/Tinkoff/maskito/issues/38

waterplea (approved):

PR #39, opened 2022-12-14
feat(core): new `maskitoPipe` utility

nsbarsukov:
https://github.com/Tinkoff/maskito/blob/43dea21494c9463bb761f31e536b85c801b6431f/projects/kit/src/lib/time/time-mask.ts#L26-L42


splincode (approved):

projects/core/src/lib/utils/pipe.ts
```diff
@@ -0,0 +1,13 @@
+import {MaskPostprocessor, MaskPreprocessor} from '../types';
+
+export function maskitoPipe(...processors: readonly MaskPreprocessor[]): MaskPreprocessor;
+
+export function maskitoPipe(
+    ...processors: readonly MaskPostprocessor[]
+): MaskPostprocessor;
+
+export function maskitoPipe(...processors: readonly Function[]): Function {
+    return (initialData: object) => {
```
waterplea, line 10: Arrow function maybe?

waterplea (approved):

projects/core/src/lib/utils/pipe.ts
nsbarsukov: ✅

PR #96, opened 2023-02-06
feat(kit): `DateRange` add `minLength` / `maxLength` options

nsbarsukov:
## PR Checklist

- [X] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/)
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Refactoring
- [ ] Code style update

**Partially** solves https://github.com/Tinkoff/maskito/issues/75

- [ ] Yes
- [X] No

### DIFF
--- a/projects/kit/src/lib/utils/is-empty.ts
+++ b/projects/kit/src/lib/utils/is-empty.ts
@@ -1,5 +1,3 @@
-export function isEmpty(
-    entity: boolean | number | object | string | null | undefined,
-): boolean {
+export function isEmpty(entity: object | null | undefined): boolean {
     return !entity || (typeof entity === 'object' && Object.keys(entity).length === 0);
 }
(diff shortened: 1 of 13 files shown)


splincode (approved):

projects/kit/src/lib/utils/is-empty.ts
```diff
@@ -0,0 +1,5 @@
+export function isEmpty(
```
splincode, line 1: is this the expected behavior?

image

nsbarsukov: It is falsy value, i expect it to return true.

Do you think I should remove primitives from the function signature ? 🤔 

```ts
export function isEmpty(
    entity: object | null | undefined,
): boolean {
```

nsbarsukov: fixed ✅

PR #183, opened 2023-03-09
feat(core): improve typing

RomanBorisov:
## PR Checklist

- [ ] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [x] Code style update

Closes # 

- [ ] Yes
- [ ] No

### DIFF
--- a/projects/core/src/lib/types/keyboard-keys.ts
+++ b/projects/core/src/lib/types/keyboard-keys.ts
@@ -1,0 +1,7 @@
+export enum KeyboardKeys {
+    Backspace = 'Backspace',
+    Delete = 'Delete',
+    Enter = 'Enter',
+    Z = 'z',
+    Y = 'y',
+}
--- a/projects/core/src/lib/types/mask-processors.ts
+++ b/projects/core/src/lib/types/mask-processors.ts
@@ -5,7 +5,7 @@ export type MaskPreprocessor = (
         elementState: ElementState;
         data: string;
     },
-    actionType: 'deleteBackward' | 'deleteForward' | 'insert' | 'validation',
+    actionType: MaskPreprocessorAction,
 ) => {
     elementState: ElementState;
     data?: string;
@@ -15,3 +15,10 @@ export type MaskPostprocessor = (
     elementState: ElementState,
     initialElementState: ElementState,
 ) => ElementState;
+
+export enum MaskPreprocessorAction {
+    DeleteBackward = 'deleteBackward',
+    DeleteForward = 'deleteForward',
+    Insert = 'insert',
+    Validation = 'validation',
+}
--- a/projects/core/src/lib/utils/test/get-not-empty-selection.spec.ts
+++ b/projects/core/src/lib/utils/test/get-not-empty-selection.spec.ts
@@ -1,0 +1,76 @@
+import {ElementState} from '../../types';
+import {getNotEmptySelection} from '../get-not-empty-selection';
+
+describe('getNotEmptySelection', () => {
+    it('should return the same selection when selection positions are not equal', () => {
+        const elementStateStub: ElementState = {
+            value: 'testValue',
+            selection: [1, 3],
+        };
+
+        expect(getNotEmptySelection(elementStateStub, true)).toEqual(
+            elementStateStub.selection,
+        );
+        expect(getNotEmptySelection(elementStateStub, false)).toEqual(
+            elementStateStub.selection,
+        );
+    });
+
+    describe('backward direction', () => {
+        it('should decrease by one start position value', () => {
+            const elementStateStub: ElementState = {
+                value: 'testValue',
+                selection: [4, 4],
+            };
+
+            expect(getNotEmptySelection(elementStateStub, false)).toEqual([3, 4]);
+        });
+
+        it('should not change everything when start value is 0', () => {
+            const elementStateStub: ElementState = {
+                value: 'testValue',
+                selection: [0, 4],
+            };
+
+            expect(getNotEmptySelection(elementStateStub, false)).toEqual([0, 4]);
+        });
+
+        it('should decrease by one start value', () => {
+            const elementStateStub: ElementState = {
+                value: 'testValue',
+                selection: [1, 1],
+            };
+
+            expect(getNotEmptySelection(elementStateStub, false)).toEqual([0, 1]);
+        });
+    });
+
+    describe('forward direction', () => {
+        it('should increase by one end position, when value`s length more then end position', () => {
+            const elementStateStub: ElementState = {
+                value: 'testValue',
+                selection: [2, 2],
+            };
+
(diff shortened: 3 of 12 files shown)


projects/core/src/lib/types/keyboard-keys.ts
```diff
@@ -0,0 +1,7 @@
+export enum KeyboardKeys {
```
nsbarsukov, line 1: I didn't see any usage of this new enum.

Do we really need it ? 🤔

projects/core/src/lib/utils/test/get-not-empty-selection.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+            expect(getNotEmptySelection(elementStateStub, true)).toEqual([2, 3]);
+        });
+
+        it('should return value length as end position, when value`s length less or equal then end position ', () => {
+            const elementStateStub: ElementState = {
+                value: 'sx',
+                selection: [4, 4],
+            };
+
+            expect(getNotEmptySelection(elementStateStub, true)).toEqual([4, 2]);
```
nsbarsukov, line 64: Good catch.
I think that we should change `getNotEmptySelection` little bit 🤔 

```ts
export function getNotEmptySelection(
    {value, selection}: ElementState,
    isForward: boolean,
): SelectionRange {
    const [from, to] = selection;

    if (from !== to) {
        return [from, to];
    }

    const notEmptySelection = isForward ? [from, to + 1] : [from - 1, to];

    return notEmptySelection.map(x => Math.min(Math.max(x, 0), value.length)) as [
        number,
        number,
    ];
}
```

projects/core/src/lib/types/mask-processors.ts
```diff
@@ -15,3 +15,10 @@ export type MaskPostprocessor = (
     elementState: ElementState,
     initialElementState: ElementState,
 ) => ElementState;
+
+export enum MaskPreprocessorAction {
```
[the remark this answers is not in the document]
nsbarsukov: (+ move this enum to the separate file)

projects/core/src/lib/types/mask-processors.ts
```diff
@@ -5,7 +5,7 @@ export type MaskPreprocessor = (
         elementState: ElementState;
         data: string;
     },
-    actionType: 'deleteBackward' | 'deleteForward' | 'insert' | 'validation',
+    actionType: MaskPreprocessorAction,
```
nsbarsukov, line 8: String union type has equal strength of typing as enum (without increasing library' bundle as enum does).

@RomanBorisov 
Do you propose this change because you want to DX ? 🤔

nsbarsukov: @waterplea @vladimirpotekhin @splincode 

What do you think about it? Could we afford to increase bundle by extra object ? :)

nsbarsukov:
@RomanBorisov Thanks for you contribution ❤️

projects/core/src/lib/types/mask-processors.ts
RomanBorisov: If main problem is bundle size, I can rewrite enum defenition like this
`export const enum MaskPreprocessorAction`
When we use const target enum does not appear in prod bundle. Example 
https://www.typescriptlang.org/play?#code/MYewdgzgLgBApmArgWxgZQJbIA4Bs4CiSqA3jAFAxUwCCMAvDAOQCGTANJdQEIPMBGHLlQDCfJsCbkAvuXL5YLAFzoseQsT6Yc+IigB0NIA

package.json
```diff
@@ -38,6 +38,8 @@
         "test": "nx test",
         "*** Format ***": "",
         "lint": "eslint .",
+        "lint:fix": "eslint . --fix",
```
splincode, line 41: if you want fix you can use  `npm run lint -- --fix` instead `npm run lint:fix`

projects/core/src/lib/types/keyboard-keys.ts
```diff
@@ -0,0 +1,7 @@
+export const enum KeyboardKeys {
```
splincode, line 1: But now we have a new object in bundle

image


because we use `preserveConstEnums` for compile libs

https://www.typescriptlang.org/play?preserveConstEnums=true#code/KYDwDg9gTgLgBAYwgOwM72MgrgWzgaWAE8AjCAQygBNCjU4BvAKDlbgCFyEBrVML4HAC8cAOScefAaIA0LNgBFgAG2AxBI0UtXrZ81gFFk6qMLFGTetnABaZ0QC8rbAJr2iegL5A

projects/core/src/lib/types/maskito-preprocessor-action.ts
```diff
@@ -0,0 +1,6 @@
+export const enum MaskitoPreprocessorAction {
```
splincode, line 1: the same

projects/core/src/lib/types/mask-processors.ts
vladimirpotekhin: > I can rewrite enum defenition like this

We can't export `const enum` from lib, unfortunately  https://ncjamieson.com/dont-export-const-enums/

> Could we afford to increase bundle by extra object ? :)

I think this particular object will not noticeably affect the weight, but I would prefer to use [only union type](https://www.bam.tech/article/should-you-use-enums-or-union-types-in-typescript) in the project, so that everything is uniform and not to bother with this in the future

nsbarsukov: @RomanBorisov We discussed this moment with the team.

Let's revert changes with `MaskitoPreprocessorAction` enum and keep union type (as it was).

Other changes is useful, let's keep them:
* projects/core/src/lib/utils/get-not-empty-selection.ts ✅ 
* projects/core/src/lib/utils/test/get-not-empty-selection.spec.ts ✅ 

___

About `KeyboardKeys` enum 🤔 
I think it is better to improve typings in same way as `TypedInputEvent`:
https://github.com/Tinkoff/maskito/blob/ff48b73926205c36c78d88f0c1d1fc320f054533/projects/core/src/lib/types/typed-input-event.ts#L1-L4

We can introduce `TypedKeyboardEvent`.
But let's do it for separate PR.

nsbarsukov:
@RomanBorisov
Could you update your PR ?

Changes from the following files are useful:
* projects/core/src/lib/utils/get-not-empty-selection.ts ✅
* projects/core/src/lib/utils/test/get-not-empty-selection.spec.ts ✅

Could you keep only them them?

RomanBorisov:
@nsbarsukov I'm sorry, there was some issues with git. I created a new PR https://github.com/Tinkoff/maskito/pull/215. Hope it's not problem?

Issue #310, opened 2023-05-19
🐞 - ⌘⇧Z stops working if the layout is not English


demensky:
### Which package(s) are the source of the bug?

@maskito/core

### Playground Link

https://tinkoff.github.io/maskito/recipes/phone#us

### Description

https://github.com/Tinkoff/maskito/assets/10235949/88d2f238-f5e0-4b62-9b47-16b360578aa0

### Maskito version

0.12.0

### Which browsers have you used?

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
**TIL**

Use english keyboard + press `Cmd` + `Shift` + `Z` => `keyboardEvent.key` equal to `z`.
Use NOT english keyboard + press `Cmd` + `Shift` + `Z` => `keyboardEvent.key` equal to `Z` (uppercase).

😄

nsbarsukov:
@demensky Do you want to become contributor of this project ?

The only required change is this line:
https://github.com/Tinkoff/maskito/blob/b6e822859ba233689dbeccd24d290f29ca190b97/projects/core/src/lib/mask.ts#L38

Something like:

```
key.toLowerCase()
```

or  use [KeyboardEvent: code property](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code).

nsbarsukov:
I think that [KeyboardEvent: code property](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) is the best solution 🤔

waterplea:
> I think that [KeyboardEvent: code property](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) is the best solution 🤔

Just gotta make sure we test it with different keyboard layouts.

demensky:
Okay, @nsbarsukov, I'll do a PR today tonight.

nsbarsukov:
This issue is not solved completely.

Further improvements will be done in this issue:
* https://github.com/Tinkoff/maskito/issues/315

PR #313, opened 2023-05-19
fix(core): a layout-independent way to detect undo and redo shortcuts

demensky:
A more reliable way to determine the pressed key is used (using [KeyboardEvent#code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)).

Also added handling of `Ctrl` + `Shift` + `Z` on Windows.

Closes: #310

### DIFF
--- a/projects/core/src/lib/mask.ts
+++ b/projects/core/src/lib/mask.ts
@@ -11,6 +11,7 @@ import {
     isEventProducingCharacter,
     maskitoTransform,
 } from './utils';
+import {isRedo, isUndo} from './utils/dom/history-events';
 
 export class Maskito extends MaskHistory {
     private readonly isTextArea = this.element.nodeName === 'TEXTAREA';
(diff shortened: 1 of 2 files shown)


waterplea (approved):

projects/core/src/lib/mask.ts
```diff
@@ -12,6 +12,21 @@ import {
...
 
+function isRedo({ctrlKey, metaKey, shiftKey, code}: KeyboardEvent): boolean {
+    return (
+        (ctrlKey && code === 'KeyY') || // Windows
+        (ctrlKey && shiftKey && code === 'KeyZ') || // Windows & Android
+        (metaKey && shiftKey && code === 'KeyZ') // macOS & iOS
+    );
+}
+
+function isUndo({ctrlKey, metaKey, code}: KeyboardEvent): boolean {
```
nsbarsukov, line 23: Thanks for your contribution!

Could you move both new utils to `projects/core/src/lib/utils/dom/history-events.ts`, please ?

nsbarsukov (approved):

nsbarsukov (approved):

PR #316, opened 2023-05-23
feat(angular): `[maskitoElement]` can accept asynchronous predicate

### LINKED ISSUE #300: 🚀 - Allow for async maskito element predicate

@maskito/angular, @maskito/react

### Description

The predicate behavior for `maskitoElement` is currently synchronous. 

For example:

```html
<my-component [maskito]="maskitoOptions" [maskitoElement]="predicate"></my-component>
```

Where `predicate` is typically querying the desired input/textarea node:
```ts
readonly predicate: MaskitoElementPredicate = (element) => element.querySelector('input')!;
```

This works great in a variety of scenarios, but is currently not compatible with Stencil hydrated web components and I would suspect could be problematic in scenarios where the `input`/`textarea` is rendered after the host is mounted.

If `MaskitoElementPredicate` additionally supported `Promise<HTMLInputElement | HTMLTextareaElement>`, then these scenarios could be supported.

### Use cases

1. Ionic Framework (Angular)

In the Angular implementation of Ionic Framework, we use Stencil web components. The template has not mounted at the time of `ngOnChanges` firing in the `MaskitoDirective`.

Desired behavior:

```html
<ion-input [maskito]="maskitoOptions" [maskitoElement]="predicate"></ion-input>
```

```ts
readonly predicate: MaskitoElementPredicate = async element => await element.getInputElement();
```

> `getInputElement()` is a [method on the web component](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/input/input.tsx#L402-L405) that queries and returns the native `input` element. 

Repo of problematic behavior: https://stackblitz.com/edit/angular-gvhxc7?file=package.json,src%2Fapp%2Fexample.component.ts,src%2Fapp%2Fexample.component.html

You can observe the mask does not initially load. Interacting in the editor will cause the iframe to reload and the mask will work as expected. 

### Alternatives

1. Ionic Framework (Angular)

We can manually invoke a `new Maskito` instance from `@maskito/core` when we return the result from our web component's API. This however is a more involved approach for Angular developers.

### Additional Information

Let me know if this is a feature you would consider. I'm happy to contribute the work involved to support this if so.

sean-perkins:
Let me know if this is an undesirable approach to this problem. I did confirm with installing `@ionic/angular@6` into the demo app (had to use v6 due to Angular 12 dependencies) that an asynchronous predicate would solve the compatibility issues with using Ionic web components with Maskito via the directive. I opted against an `Observable` return type, since `rxjs` is not a dependency of `@maskito/core`.

Outstanding questions:
- What is the best way to confirm changes against the React package? There does not appear to be a demo application for React.

## PR Checklist

- [x] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/)
- [x] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

The `maskitoElement` is synchronous, _only_ allowing a developer to query a node from the host element (`querySelector`, etc.).

Closes #300

## What is the new behavior?

The `maskitoElement` is asynchronous, allowing a developer to perform asynchronous actions to resolve the node from the host element.

For example in Ionic Framework, where the native `input` element is resolved from a shadow DOM web component:
```ts
readonly asyncPredicate: MaskitoElementPredicate<HTMLIonInputElement> = async (element) => element.getInputElement();
```

- [ ] Yes
- [x] No

This change is backwards compatible with the existing type signature of resolving an node that is either `HTMLInputElement` or `HTMLTextareaElement`.

### DIFF
--- a/projects/angular/src/lib/maskito.directive.ts
+++ b/projects/angular/src/lib/maskito.directive.ts
@@ -12,6 +12,7 @@ import {
     MASKITO_DEFAULT_ELEMENT_PREDICATE,
     MASKITO_DEFAULT_OPTIONS,
     MaskitoElementPredicate,
+    MaskitoElementPredicateAsync,
     MaskitoOptions,
 } from '@maskito/core';
 
@@ -25,7 +26,8 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
     maskito: MaskitoOptions = MASKITO_DEFAULT_OPTIONS;
 
     @Input()
-    maskitoElement: MaskitoElementPredicate = MASKITO_DEFAULT_ELEMENT_PREDICATE;
+    maskitoElement: MaskitoElementPredicate | MaskitoElementPredicateAsync =
+        MASKITO_DEFAULT_ELEMENT_PREDICATE;
 
     constructor(
         @Inject(NgZone) private readonly ngZone: NgZone,
--- a/projects/angular/src/lib/maskito.directive.ts
+++ b/projects/angular/src/lib/maskito.directive.ts
@@ -11,6 +11,7 @@ import {
     Maskito,
     MASKITO_DEFAULT_ELEMENT_PREDICATE,
     MASKITO_DEFAULT_OPTIONS,
+    MaskitoElementPredicate,
     MaskitoOptions,
 } from '@maskito/core';
 
@@ -24,7 +25,7 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
     maskito: MaskitoOptions = MASKITO_DEFAULT_OPTIONS;
 
     @Input()
-    maskitoElement = MASKITO_DEFAULT_ELEMENT_PREDICATE;
+    maskitoElement: MaskitoElementPredicate = MASKITO_DEFAULT_ELEMENT_PREDICATE;
 
     constructor(
         @Inject(NgZone) private readonly ngZone: NgZone,
@@ -44,7 +45,7 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
         }
 
         this.ngZone.runOutsideAngular(() => {
-            this.maskedElement = new Maskito(predicateResult, this.maskito,);
+            this.maskedElement = new Maskito(predicateResult, this.maskito);
         });
     }
 
--- a/projects/angular/src/lib/maskito.directive.ts
+++ b/projects/angular/src/lib/maskito.directive.ts
@@ -31,13 +31,20 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
         @Inject(ElementRef) private readonly elementRef: ElementRef<HTMLElement>,
     ) {}
 
-    ngOnChanges(): void {
+    async ngOnChanges(): Promise<void> {
         this.maskedElement?.destroy();
+
+        const predicate = this.maskitoElement;
+        const predicateResult = await predicate(this.elementRef.nativeElement);
+
+        if (this.maskitoElement !== predicate) {
+            // Ignore the result of the predicate if the
+            // maskito element has changed before the predicate was resolved.
+            return;
+        }
+
         this.ngZone.runOutsideAngular(() => {
-            this.maskedElement = new Maskito(
-                this.maskitoElement(this.elementRef.nativeElement),
-                this.maskito,
-            );
+            this.maskedElement = new Maskito(predicateResult, this.maskito,);
         });
     }
 
--- a/projects/core/src/index.ts
+++ b/projects/core/src/index.ts
@@ -6,7 +6,6 @@ export {Maskito} from './lib/mask';
 export {
     MaskitoElementPredicate,
     MaskitoElementPredicateAsync,
-    MaskitoElementPredicateSync,
     MaskitoMask,
     MaskitoOptions,
     MaskitoPlugin,
--- a/projects/core/src/index.ts
+++ b/projects/core/src/index.ts
@@ -4,11 +4,13 @@ export {
 } from './lib/constants';
 export {Maskito} from './lib/mask';
 export {
+    AsyncMaskitoElementPredicate,
     MaskitoElementPredicate,
     MaskitoMask,
     MaskitoOptions,
     MaskitoPlugin,
     MaskitoPostprocessor,
     MaskitoPreprocessor,
+    SyncMaskitoElementPredicate,
 } from './lib/types';
 export {maskitoPipe, maskitoTransform} from './lib/utils';
--- a/projects/core/src/index.ts
+++ b/projects/core/src/index.ts
@@ -4,13 +4,13 @@ export {
 } from './lib/constants';
 export {Maskito} from './lib/mask';
 export {
-    AsyncMaskitoElementPredicate,
     MaskitoElementPredicate,
+    MaskitoElementPredicateAsync,
+    MaskitoElementPredicateSync,
     MaskitoMask,
     MaskitoOptions,
     MaskitoPlugin,
     MaskitoPostprocessor,
     MaskitoPreprocessor,
-    SyncMaskitoElementPredicate,
 } from './lib/types';
 export {maskitoPipe, maskitoTransform} from './lib/utils';
--- a/projects/core/src/lib/constants/default-element-predicate.ts
+++ b/projects/core/src/lib/constants/default-element-predicate.ts
@@ -1,5 +1,4 @@
 import {MaskitoElementPredicate} from '../types';
 
 export const MASKITO_DEFAULT_ELEMENT_PREDICATE: MaskitoElementPredicate = e =>
-    e.querySelector<HTMLInputElement | HTMLTextAreaElement>('input,textarea') ||
-    (e as HTMLInputElement | HTMLTextAreaElement);
+    e.querySelector('input,textarea') || (e as HTMLInputElement | HTMLTextAreaElement);
--- a/projects/core/src/lib/constants/default-element-predicate.ts
+++ b/projects/core/src/lib/constants/default-element-predicate.ts
@@ -1,5 +1,5 @@
-import {SyncMaskitoElementPredicate} from '../types';
+import {MaskitoElementPredicateSync} from '../types';
 
-export const MASKITO_DEFAULT_ELEMENT_PREDICATE: SyncMaskitoElementPredicate = e =>
+export const MASKITO_DEFAULT_ELEMENT_PREDICATE: MaskitoElementPredicateSync = e =>
     e.querySelector<HTMLInputElement | HTMLTextAreaElement>('input,textarea') ||
     (e as HTMLInputElement | HTMLTextAreaElement);
--- a/projects/core/src/lib/constants/default-element-predicate.ts
+++ b/projects/core/src/lib/constants/default-element-predicate.ts
@@ -1,5 +1,5 @@
-import {MaskitoElementPredicateSync} from '../types';
+import {MaskitoElementPredicate} from '../types';
 
-export const MASKITO_DEFAULT_ELEMENT_PREDICATE: MaskitoElementPredicateSync = e =>
+export const MASKITO_DEFAULT_ELEMENT_PREDICATE: MaskitoElementPredicate = e =>
     e.querySelector<HTMLInputElement | HTMLTextAreaElement>('input,textarea') ||
     (e as HTMLInputElement | HTMLTextAreaElement);
--- a/projects/core/src/lib/constants/default-element-predicate.ts
+++ b/projects/core/src/lib/constants/default-element-predicate.ts
@@ -1,4 +1,5 @@
-import {MaskitoElementPredicate} from '../types';
+import {SyncMaskitoElementPredicate} from '../types';
 
-export const MASKITO_DEFAULT_ELEMENT_PREDICATE: MaskitoElementPredicate = e =>
-    e.querySelector('input,textarea') || (e as HTMLInputElement | HTMLTextAreaElement);
+export const MASKITO_DEFAULT_ELEMENT_PREDICATE: SyncMaskitoElementPredicate = e =>
+    e.querySelector<HTMLInputElement | HTMLTextAreaElement>('input,textarea') ||
+    (e as HTMLInputElement | HTMLTextAreaElement);
--- a/projects/demo-integrations/cypress/tests/angular/predicate.cy.ts
+++ b/projects/demo-integrations/cypress/tests/angular/predicate.cy.ts
@@ -13,4 +13,16 @@ describe('@maskito/angular | Predicate', () => {
 
         cy.get('@name').focus().type('12341234abcd12341234').should('have.value', 'ABCD');
     });
+    it('supports asynchronous predicate', () => {
+        cy.visit(DemoPath.Cypress);
+        cy.get('#predicate #async-predicate input')
+            .should('be.visible')
+            .first()
+            .as('card');
+
+        cy.get('@card')
+            .focus()
+            .type('12341234abcd12341234')
+            .should('have.value', '1234 1234 1234 1234');
+    });
 });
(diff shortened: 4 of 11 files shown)


projects/angular/src/lib/maskito.directive.ts
```diff
@@ -30,16 +30,16 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
...
-        this.ngZone.runOutsideAngular(() => {
-            this.maskedElement = new Maskito(
-                this.maskitoElement(this.elementRef.nativeElement),
-                this.maskito,
-            );
+
+        await this.ngZone.runOutsideAngular(async () => {
+            const predicate = await this.maskitoElement(this.elementRef.nativeElement);
+
+            this.maskedElement = new Maskito(predicate, this.maskito);
```
nsbarsukov, line 42: We gotta be wary of race conditions.

What happens if async predicate function will return value too long?

For example:



1. Developer pass async `predicate#1`
2. ngOnChanges fires and execute `predicate#1`
3. `predicate#1` will return value only in 5 seconds
4. Developer pass new async `predicate#2` (before `predicate#1` is complete)
5. ngOnChanges fires and execute `predicate#2`
6. `predicate#2` instantly completes and create `Mask#2`
7. `predicate#1` completes and create `Mask#1`

nsbarsukov: For `@maskito/angular` package we could wrap it with `Observable` + use `switchMap`.

projects/react/src/lib/useMaskito.ts
```diff
@@ -32,26 +33,32 @@ export const useMaskito = ({
...
+    >(null);
 
     const onRefChange: RefCallback<HTMLElement> = useCallback(
-        (node: HTMLElement | null) => {
-            setElement(node);
+        async (node: HTMLElement | null) => {
+            if (node) {
+                const predicate = await elementPredicate(node);
+
+                setPredicateResult(predicate);
```
nsbarsukov, line 45: @nextZed Could you review changes inside this file (`@maskito/react`), please ?

projects/angular/src/lib/maskito.directive.ts
```diff
@@ -30,16 +30,16 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
 
     @Input()
     maskitoElement: MaskitoElementPredicate = e =>
-        e.querySelector('input,textarea') ||
+        e.querySelector<HTMLInputElement | HTMLTextAreaElement>('input,textarea') ||
```
waterplea, line 33: This is unnecessary, type is inferred from the `MaskitoElementPredicate` type of the arrow function.

projects/angular/src/lib/maskito.directive.ts
nsbarsukov: I don't know what we should do for `@maskito/react`.

@nextZed Could you suggest solution to process this case in **React** way ?

nsbarsukov (commented):
Thanks for your contribution!

projects/angular/src/lib/maskito.directive.ts
```diff
@@ -30,16 +30,16 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
...
+    async ngOnChanges(): Promise<void> {
         this.maskedElement?.destroy();
-        this.ngZone.runOutsideAngular(() => {
-            this.maskedElement = new Maskito(
-                this.maskitoElement(this.elementRef.nativeElement),
-                this.maskito,
-            );
+
+        await this.ngZone.runOutsideAngular(async () => {
+            const predicate = await this.maskitoElement(this.elementRef.nativeElement);
```
waterplea, line 40: I think this should be outside ngZone wrap:
```ts
const element = await this.maskitoElement(this.elementRef.nativeElement);

this.ngZone.runOutsideAngular(() => {
    this.maskedElement = new Maskito(predicate, this.maskito);
});
```
This way asynchronous part is more pinpoint. If the predicate is synchronous, does this introduce a microtask delay to the processing or does it work synchronously? I don't really remember the specs on this, but I think it is synchronous in that case 🤔

projects/demo/src/pages/documentation/js-frameworks-angular/examples/1-nested/component.ts
```diff
@@ -15,5 +15,5 @@ export class NestedDocExample1 {
     };
 
     readonly predicate: MaskitoElementPredicate = element =>
-        element.querySelector('tui-input input')!;
+        element.querySelector<HTMLInputElement>('tui-input input')!;
```
waterplea, line 18: I think this doesn't really do anything, because this arrow function is already typed as `MaskitoElementPredicate`, and narrowing it down from `HTMLInputElement | HTMLTextAreaElement` is irrelecant.

waterplea (commented):
I am worried about race conditions. What if we set one predicate and before the promise is resolved change it to another that resolved and then we get result from the previous predicate. Is this a possibility?

projects/angular/src/lib/maskito.directive.ts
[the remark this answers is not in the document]
waterplea: I didn't see your comments before starting my review. I think it's good to have a constant, I would just suggest to move it out of `ngZone` wrap so we do not have to make that asynchronous. And yeah, renaming it to `predicateResult` like in React lib would be more suitable.

projects/demo-integrations/cypress/tests/angular/predicate.cy.ts
```diff
@@ -13,4 +13,16 @@ describe('@maskito/angular | Predicate', () => {
 
         cy.get('@name').focus().type('12341234abcd12341234').should('have.value', 'ABCD');
```
nsbarsukov, line 14: This old test fails.

See [job's logs](https://github.com/Tinkoff/maskito/actions/runs/5062105791/jobs/9094853389?pr=316).

```
  @maskito/angular | Predicate
    1) can detect run-time changes
    ✓ supports asynchronous predicate (1160ms)


  1 passing (14s)
  1 failing

  1) @maskito/angular | Predicate
       can detect run-time changes:

      Timed out retrying after 10250ms
      + expected - actual

      -'1234 1234 1234 1234'
      +'ABCD'

```

nsbarsukov (changes requested):

projects/angular/src/lib/maskito.directive.ts
waterplea: I have an idea, what if we save predicate to a const before resolving it and then check if it remained the same? Pseudo code:
```ts
const predicate = this.predicate
const result = async predicate(element);

if (this.predicate !== predicate) {
  return;
}

this.maskedElement = new Maskito(result, options);
```

projects/react/src/lib/useMaskito.ts
nextZed: I think we can't make ref callback async, because ref callbacks run synchronously in commit phase. I suggest to separate predicates typings 
```ts
export type SyncMaskitoElementPredicate<T extends HTMLElement = HTMLElement> = (
    element: T,
) =>
    | HTMLInputElement
    | HTMLTextAreaElement

export type AsyncMaskitoElementPredicate<T extends HTMLElement = HTMLElement> = (
    element: T,
) => Promise<HTMLInputElement | HTMLTextAreaElement>;

export type MaskitoElementPredicate<T extends HTMLElement = HTMLElement> = SyncMaskitoElementPredicate<T> | AsyncMaskitoElementPredicate<T>
``` 

and use only `SyncMaskitoElementPredicate` in `@maskito/react`

projects/react/src/lib/useMaskito.ts
```diff
@@ -32,26 +33,32 @@ export const useMaskito = ({
...
+            }
         },
-        [],
+        [elementPredicate],
```
nextZed, line 48: I think it is better to make ref callback dumb as possible and without any dependencies

in your case if user changes predicate on the fly it won't automatically update

projects/react/src/lib/useMaskito.ts
```diff
@@ -32,26 +33,32 @@ export const useMaskito = ({
...
+    const [predicateResult, setPredicateResult] = useState<
+        HTMLInputElement | HTMLTextAreaElement | null
+    >(null);
 
     const onRefChange: RefCallback<HTMLElement> = useCallback(
-        (node: HTMLElement | null) => {
-            setElement(node);
+        async (node: HTMLElement | null) => {
+            if (node) {
+                const predicate = await elementPredicate(node);
```
nsbarsukov, line 43: @sean-perkins 
Could you add new unit tests (to test async predicate) for `@maskito/react` package, please ?

https://github.com/Tinkoff/maskito/blob/5783e766a657abcf0fc7f8a8d12ac1bf412dc18a/projects/react/src/lib/useMaskito.spec.tsx#L22

projects/demo/src/pages/documentation/js-frameworks-angular/examples/1-nested/component.ts
sean-perkins: I'll research this further, once the `Promise` is added to the `MaskitoElementPredicate` Typescript produces this error:
```
Type 'Element' is not assignable to type 'HTMLInputElement | HTMLTextAreaElement | Promise<HTMLInputElement | HTMLTextAreaElement>'.

Type 'Element' is missing the following properties from type 'HTMLTextAreaElement': autocomplete, cols, defaultValue, dirName, and 137 more.ts
```

This applies to: https://github.com/Tinkoff/maskito/pull/316#discussion_r1203563251 as well. 

This may be a weird quirk or bug in Typescript. I'm also observing that if the selector is updated to:

```ts
element.querySelector('input')!;
```

The error goes away and the inferred type is `HTMLInputElement`. With `tui-input input`, the inferred type is `Element`. 

If the `Promise` type is removed from `MaskitoElementPredicate`, using `tui-input input` results in the inferred type of `HTMLInputElement`.

projects/angular/src/lib/maskito.directive.ts
sean-perkins: Addressed here: https://github.com/Tinkoff/maskito/pull/316/commits/4dea275d78d7d2d86c026056df8c09a6096aaa84 (removed the VSCode settings file in a later commit). 

Based on the feedback here: https://github.com/Tinkoff/maskito/pull/316#discussion_r1203629086 I implemented this in the Angular implementation only and reverted the react changes to remain synchronous. I'll do more discovery into the usage impact with my use case of Ionic React (web components) and the Maskito React hook.

projects/core/src/index.ts
```diff
@@ -1,11 +1,13 @@
...
 export {Maskito} from './lib/mask';
 export {
+    AsyncMaskitoElementPredicate,
     MaskitoElementPredicate,
     MaskitoMask,
     MaskitoOptions,
     MaskitoPlugin,
     MaskitoPostprocessor,
     MaskitoPreprocessor,
+    SyncMaskitoElementPredicate,
```
nsbarsukov, line 11: All public API should be start from `maskito` / `Maskito` prefix. 

Let's rename new types:
```
SyncMaskitoElementPredicate => MaskitoElementPredicateSync
AsyncMaskitoElementPredicate => MaskitoElementPredicateAsync
```

nextZed:
@sean-perkins does ionic need maskito element predicate to be async in ionic/react?
Just letting you know that in this https://github.com/Tinkoff/maskito/pull/316#discussion_r1203629086 I'm afraid of race condition. Code that you've implemented yesterday will work, but there is a chance, that bad use of it can lead to unexpected behaviour.

If ionic/react needs it, we can implement that and handle some way.

waterplea:
Hi @sean-perkins, I know ionic also has Vue, if you rebase your branch - you should be able to easily add async predicate to Vue package in the same manner it is done in Angular. You would probably need to add another global map to keep track of predicates though. Don't forget to clean it up in unmount hook.

sean-perkins:
Ok, I've finally gotten the time to test the latest changes against each output of Ionic for Angular, React and Vue. 

The async predicate behavior is only beneficial to Ionic Angular, since we are using what is called a "lazy" output of Stencil web components that initialize as they are added to the DOM (this process can cause a few ms delay to the template mounting). We do have plans at Ionic to move to web custom elements output, which would not require the the async predicate behavior in the future.

For both React and Vue, Ionic uses the web custom elements output and the async predicate behavior is not needed. 

With the current state of the PR, everything does work great across our official Ionic integrations with frameworks. 

Knowing this - how would you all like to continue? It feels strange to ask you to add async predicate support to Angular for this narrow use case. It is possible I can come up with an implementation workaround for Angular.

waterplea:
@nsbarsukov what do you think? I believe we should support async predicate in Angular, because it can be useful in sync cases too — if input is hidden under *ngIf inside custom component, even if it's truthy, it would synchronously appear in DOM after maskito inputs are processed. We had this issue with an old mask in Taiga when I added *ngIf just to declare some Observable value with async pipe — and mask did not find the input.

nsbarsukov:
Another my opinion is to add async predicate to all frameworks.
I like when all packages have the similar behaviour.

But I don't understand situation with React package.
Do we have any arguments against async predicate inside this package ?

waterplea:
> Another my opinion is to add async predicate to all frameworks. I like when all packages have the similar behaviour.
> 
> But I don't understand situation with React package. Do we have any arguments against async predicate inside this package ?

If we can somehow add it to React in a reliable way, it would be the best solution. Adding async predicate to Vue should be as easy as with Angular.

vladimirpotekhin:
> I am not against to add async predicate for Angular only.
> My idea is keep `MaskitoElementPredicate` as it is now (inside `@maskito/core`). And create `MaskitoElementPredicateAsync` only.

I like this @nbarsukov idea. For now we can add async predicate `MaskitoElementPredicateAsync` for Angular only. And later we will add to other frameworks when the need arises. Once all frameworks have an asynchronous predicate, we can export `MaskitoElementPredicateSync = MaskitoElementPredicate`

waterplea:
@sean-perkins let's add it to Angular package only at this time so that at least this PR can be merged and then we will think about other frameworks. Let's keep the current name in core package and inline Promise based type in Angular directive input for now.

nsbarsukov:
@sean-perkins let's put both `MaskitoElementPredicate` and `MaskitoElementPredicateAsync` inside the same file `projects/core/src/lib/types/element-predicate.ts` (`@maskito/core`), yep

projects/core/src/lib/constants/default-element-predicate.ts
```diff
@@ -1,4 +1,5 @@
 import {MaskitoElementPredicate} from '../types';
 
 export const MASKITO_DEFAULT_ELEMENT_PREDICATE: MaskitoElementPredicate = e =>
-    e.querySelector('input,textarea') || (e as HTMLInputElement | HTMLTextAreaElement);
+    e.querySelector<HTMLInputElement | HTMLTextAreaElement>('input,textarea') ||
```
nsbarsukov, line 4: Could we revert these lines ? 

See https://github.com/Tinkoff/maskito/pull/316#discussion_r1203573546

projects/demo/src/pages/frameworks/angular/examples/1-nested/component.ts
```diff
@@ -15,5 +15,5 @@ export class NestedDocExample1 {
     };
 
     readonly predicate: MaskitoElementPredicate = element =>
-        element.querySelector('tui-input input')!;
+        element.querySelector<HTMLInputElement>('tui-input input')!;
```
nsbarsukov, line 18: https://github.com/Tinkoff/maskito/pull/316#discussion_r1203573546

projects/core/src/lib/constants/default-element-predicate.ts
sean-perkins: Yes 👍 with the recent type changes, we can revert these. 

Addressed here: https://github.com/Tinkoff/maskito/pull/316/commits/e2b89eb84b7a5b1d483b287e45b7197466aae577

nsbarsukov (approved):

nsbarsukov:
@sean-perkins Thanks for your contribution 👍

Issue #315, opened 2023-05-23
🚀 - Find better layout-independent way to detect `Undo` and `Redo` shortcuts


nsbarsukov:
### Which package(s) are relevant/related to the feature request?

@maskito/core

### Problem description
**Maskito** uses [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) property to detect undo/redo shortcuts.

https://github.com/Tinkoff/maskito/blob/5783e766a657abcf0fc7f8a8d12ac1bf412dc18a/projects/core/src/lib/utils/dom/history-events.ts#L1-L14

It works for some keyboard layouts (for example, English).
But not all languages have the same position of these keys.

For example, in German keyboard keys  `Z` and `Y` are swapped.
<img width="714" alt="Screenshot 2023-05-23 at 12 07 49" src="https://github.com/Tinkoff/maskito/assets/35179038/0eb84310-6dd4-4974-9287-c7f836aba5ce">
And the current solution does not properly work for German keyboard.

### Proposed solutions
We need investigate this issue and look for better alternatives.
For example, we can use legacy [KeyboardEvent.keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) property.

demensky:
## Undo hotkey

| Layout  | Browser | OS      | `code`  | `key` | `keyCode` | `which` |
| ------- | ------- | ------- | :-----: | :---: | :-------: | :-----: |
| English | Chrome  | macOS   | `KeyZ`  |  `z`  |   `90`    |  `90`   |
| English | Chrome  | Windows | `KeyZ`  |  `z`  |   `90`    |  `90`   |
| English | Chrome  | Linux   |    ?    |   ?   |     ?     |    ?    |
| English | Chrome  | Android | `KeyZ`  |  `z`  |   `90`    |  `90`   |
| English | Safari  | macOS   | `KeyZ`  |  `z`  |   `90`    |  `90`   |
| English | Firefox | macOS   | `KeyZ`  |  `z`  |   `90`    |  `90`   |
| English | Firefox | Windows | `KeyZ`  |  `z`  |   `90`    |  `90`   |
| English | Firefox | Linux   |    ?    |   ?   |     ?     |    ?    |
| English | Firefox | Android |    ?    |   ?   |     ?     |    ?    |
| Russian | Chrome  | macOS   | `KeyZ`  |  `z`  |   `90`    |  `90`   |
[...]

PR #320, opened 2023-05-25
feat(core): better layout-independent way to detect Undo and Redo

demensky:
Closes: #315

### DIFF
--- a/projects/core/src/lib/utils/dom/hotkey.ts
+++ b/projects/core/src/lib/utils/dom/hotkey.ts
@@ -1,8 +1,8 @@
 export const enum HotkeyModifier {
-    CTRL = 0b0001,
-    ALT = 0b0010,
-    SHIFT = 0b0100,
-    META = 0b1000,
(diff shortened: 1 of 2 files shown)


waterplea (approved):

nsbarsukov (approved):

projects/core/src/lib/utils/dom/hotkey.ts
```diff
@@ -0,0 +1,44 @@
+export const enum HotkeyModifier {
+    CTRL = 0b0001,
+    ALT = 0b0010,
+    SHIFT = 0b0100,
+    META = 0b1000,
```
nsbarsukov, line 5: Cool idea to use bit **mask** inside textfield **mask** 😄 👍 

**nitpicking:**

```suggestion
    CTRL = 1 << 0,
    ALT = 1 << 1,
    SHIFT = 1 << 2,
    META = 1 << 3,
```

It is just the different way to do the same thing but it is easier to extend enum with new values.

PR #373, opened 2023-07-18
fix(kit): `Number` should remove repeated leading zeroes for integer part only on `blur`-event

### LINKED ISSUE #370: 🐞 - `Number` trim zeroes after deleting of leading digit will break UX

@maskito/kit

### Playground Link

https://tinkoff.github.io/maskito/kit/number/API?thousandSeparator=_&precision=2

### Description

1. Open https://tinkoff.github.io/maskito/kit/number/API?thousandSeparator=_&precision=2
2. Type 1000000 (one million)
3. Want to change to 2000000 (two million), move caret after the first digit (1|_000_000)
4. Press Backspace and after press `2`.

**Expected behaviour:** `2_000_000`
Actual behaviour: `20`

1. Open https://tinkoff.github.io/maskito/kit/number/API?thousandSeparator=_&precision=2
2. Type 10023.
3. Want to change to 20023, move caret after the first digit (1|0_023).
4. Press Backspace and after press `2`.

**Expected behaviour:** `20_023`
Actual behaviour: `223`

Current behaviour is break usual UX because differs from the standard input actions.
Related issue: https://github.com/Tinkoff/maskito/issues/266.
**Workaround:** add a flag to select behavior.

### Maskito version

1.2.1

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [X] Android

nsbarsukov:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #370

### DIFF
--- a/projects/demo-integrations/cypress/tests/kit/number/number-prefix-postfix.cy.ts
+++ b/projects/demo-integrations/cypress/tests/kit/number/number-prefix-postfix.cy.ts
@@ -91,35 +91,41 @@ describe('Number | Prefix & Postfix', () => {
                 .should('have.prop', 'selectionEnd', '$0.12'.length);
         });
 
-        describe('it removes repeated leading zeroes for integer part', () => {
-            it('Type 000000 => $0| per day', () => {
+        describe('it removes repeated leading zeroes for integer part on blur', () => {
+            it('Type 000000 => blur => $0| per day', () => {
                 cy.get('@input')
                     .type('000000')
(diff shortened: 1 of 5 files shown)


waterplea (approved):
Interesting case of a "delayed processor" :) I wonder if we would need this trick again somewhere — then maybe some helper function can be made.

PR #464, opened 2023-08-31
ci: add useful workflows

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update
- [x] Build or CI related changes

### DIFF
--- a/.github/workflows/auto-label.yml
+++ b/.github/workflows/auto-label.yml
@@ -1,0 +1,14 @@
+name: 🤖 Auto label
+on: pull_request_review
+
+jobs:
+  label-when-approved:
+    name: Label when approved
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.6.0
+      - uses: taiga-family/ci/actions/variables@1.11.1
+      - uses: taiga-family/ci/actions/label-when-approved@1.11.1
+        with:
+          approvals: 1
+          token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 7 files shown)


.github/workflows/auto-remove-label.yml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  issues:
+    types: [closed]
+
+jobs:
+  remove_label:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.6.0
+      - uses: taiga-family/ci/actions/auto-remove-label@1.12.0
```
nsbarsukov, line 11: "state: need triage" label is not used inside Maskito.

I think that we can drop this workflow.

.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,13 @@
...
+on: pull_request_review
+
+jobs:
+  label-when-approved:
+    name: Label when approved
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.6.0
+      - uses: taiga-family/ci/actions/variables@1.11.1
+      - uses: taiga-family/ci/actions/label-when-approved@1.11.1
```
nsbarsukov, line 11: We need only one approve inside `Maskito` to merge PR.

Do we need this workflow ?

.github/workflows/auto-remove-label.yml
splincode: Now it's used
https://github.com/taiga-family/maskito/labels

.github/workflows/auto-label.yml
splincode: fixed

PR #605, opened 2023-10-18
feat(angular): allow nullable options

### LINKED ISSUE #387: 🚀 - Allow null in Angular directive

_No response_

### Description

Using Angular, I have a wrapped control where I want to be able to optionally pass Maskito options:

`@Input() maskito: MaskitoOptions | null = null;`

With the way the Angular directive is created, I'm not able to pass `null` as a value, so I need to use an `ngIf` to include the directive which is kind of ugly. It would be nice if `null` was a possibility and the masking would just bail early, something like:

```ts
async ngOnChanges(): Promise<void> {

    if (!this.maskito) {
        return;
    }

    const predicate = this.maskitoElement;
    const predicateResult = await predicate(this.elementRef.nativeElement);

    if (this.maskitoElement !== predicate) {
        // maskito element has changed before the predicate was resolved.
        return;
    }

    this.ngZone.runOutsideAngular(() => {
        this.maskedElement = new Maskito(predicateResult, this.maskito);
    });
}
```

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

## What is the current behaviour?

Closes #387

### DIFF
--- a/projects/angular/src/lib/maskito.cva.ts
+++ b/projects/angular/src/lib/maskito.cva.ts
@@ -21,13 +21,18 @@ import {MASKITO_DEFAULT_OPTIONS, MaskitoOptions, maskitoTransform} from '@maskit
 })
 export class MaskitoCva {
     @Input()
-    maskito: MaskitoOptions = MASKITO_DEFAULT_OPTIONS;
+    maskito: MaskitoOptions | null = MASKITO_DEFAULT_OPTIONS;
 
     constructor(readonly accessor: DefaultValueAccessor) {
         const original = accessor.writeValue.bind(accessor);
 
         accessor.writeValue = (value: unknown) => {
-            original(maskitoTransform(String(value ?? ''), this.maskito));
+            original(
+                maskitoTransform(
+                    String(value ?? ''),
+                    this.maskito ?? MASKITO_DEFAULT_OPTIONS,
+                ),
+            );
         };
     }
 }
--- a/projects/angular/src/lib/maskito.directive.ts
+++ b/projects/angular/src/lib/maskito.directive.ts
@@ -23,7 +23,7 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
     private maskedElement: Maskito | null = null;
 
     @Input()
-    maskito: MaskitoOptions = MASKITO_DEFAULT_OPTIONS;
+    maskito: MaskitoOptions | null = MASKITO_DEFAULT_OPTIONS;
 
     @Input()
     maskitoElement: MaskitoElementPredicate | MaskitoElementPredicateAsync =
@@ -47,7 +47,10 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
         }
 
         this.ngZone.runOutsideAngular(() => {
-            this.maskedElement = new Maskito(predicateResult, this.maskito);
+            this.maskedElement = new Maskito(
+                predicateResult,
+                this.maskito ?? MASKITO_DEFAULT_OPTIONS,
+            );
         });
     }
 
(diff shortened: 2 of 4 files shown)


projects/angular/src/lib/maskito.cva.ts
```diff
@@ -20,8 +20,16 @@ import {MASKITO_DEFAULT_OPTIONS, MaskitoOptions, maskitoTransform} from '@maskit
...
 export class MaskitoCva {
+    private options?: MaskitoOptions | null;
+
     @Input()
-    maskito: MaskitoOptions = MASKITO_DEFAULT_OPTIONS;
+    set maskito(options: MaskitoOptions | null | undefined) {
+        this.options = options;
+    }
+
+    get maskito(): MaskitoOptions {
```
nsbarsukov, line 30: Do we really require getter/setter ? 🤔 

Could we just do the following ?

```ts
@Input()
maskito: MaskitoOptions | null = null;

accessor.writeValue = (value: unknown) => {
            original(maskitoTransform(String(value ?? ''), this.maskito ?? MASKITO_DEFAULT_OPTIONS));
        };
```

projects/angular/src/lib/maskito.directive.ts
```diff
@@ -23,7 +23,7 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
     private maskedElement: Maskito | null = null;
 
     @Input()
-    maskito: MaskitoOptions = MASKITO_DEFAULT_OPTIONS;
+    maskito?: MaskitoOptions | null = MASKITO_DEFAULT_OPTIONS;
```
nsbarsukov, line 26: Why do we need `undefined` ?

projects/angular/src/lib/maskito.cva.ts
```diff
@@ -20,8 +20,16 @@ import {MASKITO_DEFAULT_OPTIONS, MaskitoOptions, maskitoTransform} from '@maskit
     },
 })
 export class MaskitoCva {
+    private options?: MaskitoOptions | null;
+
     @Input()
-    maskito: MaskitoOptions = MASKITO_DEFAULT_OPTIONS;
+    set maskito(options: MaskitoOptions | null | undefined) {
```
nsbarsukov, line 26: Why do we need `undefined` ?

projects/angular/src/lib/maskito.pipe.ts
```diff
@@ -5,7 +5,7 @@ import {MaskitoOptions, maskitoTransform} from '@maskito/core';
     name: 'maskito',
 })
 export class MaskitoPipe implements PipeTransform {
-    transform(value: unknown, maskitoOptions: MaskitoOptions): string {
+    transform(value: unknown, maskitoOptions?: MaskitoOptions | null): string {
```
nsbarsukov, line 8: Why do we need `undefined` ?

projects/angular/src/lib/maskito.spec.ts
```diff
@@ -58,6 +58,15 @@ describe('Maskito Angular package', () => {
...
         expect(getValue()).toBe('12345,67');
+
+        fixture.componentInstance.options = null;
+        fixture.detectChanges();
+
+        fixture.componentInstance.control.setValue(123456.9999);
+        fixture.detectChanges();
+
+        expect(getText()).toBe('123456.9999');
+        expect(getValue()).toBe('123456.9999');
```
nsbarsukov, line 69: Should it be the separate `it('disable mask formatting if options is null')` ?

projects/core/src/lib/utils/transform.ts
```diff
@@ -3,6 +3,10 @@ import {MASKITO_DEFAULT_OPTIONS} from '../constants';
 import {ElementState, MaskitoOptions} from '../types';
 import {maskitoPipe} from './pipe';
 
+export function maskitoTransform(
```
nsbarsukov, line 6: I am not sure that we need to change signature of `maskitoTransform` 🤔

projects/angular/src/lib/maskito.cva.ts
nsbarsukov: Opps, my mistake, sorry.

We should keep the link to the same object even if input prop is changed.

waterplea (approved):

nsbarsukov (approved):

PR #806, opened 2023-12-19
fix: add `{bubbles:true}` for `input` events inside all built-in plugins to support `ReactSyntheticEvent` 

### LINKED ISSUE #804: 🐞 - Plugins, that set value into input don't call onInput in React

@maskito/kit

### Playground Link

https://stackblitz.com/edit/stackblitz-starters-chezz9?description=React%20%20%20TypeScript%20starter%20project&file=src%2FApp.tsx&title=React%20Starter

### Description

Leading zero plugin and others doesn't work on blur, it sets the value, but this change don't fire in React's syntheticevent handler `onInput`, and value in state and value in input become different.

By little research i found out that it's because `bubbles` sets to false by default, and react don't handle events like this.
I've made a little example of how this works without maskito (just with react and dispatchEvent)
https://codesandbox.io/p/sandbox/pensive-shamir?file=%2Fsrc%2FApp.tsx%3A47%2C40
You can click here on buttons and it will dispatch events with and without bubbles parameter.

And example of how this works with maskito
https://stackblitz.com/edit/stackblitz-starters-chezz9?description=React%20%20%20TypeScript%20starter%20project&file=src%2FApp.tsx&title=React%20Starter
1. Open link
2. Type '000123'
3. Blur input somehow

**Expected:**
Value above input is '123'

**Actual:**
Value above input is '000 123'

To fix you can just add `bubbles: true` in plugins dispatchEvents
For example here https://github.com/taiga-family/maskito/blob/main/projects/kit/src/lib/masks/number/plugins/leading-zeroes-validation.plugin.ts#L35C29-L35C29

### Maskito version

1.9.0

- [X] Chrome
- [X] Firefox
- [X] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

aktanoff:
## PR Checklist

- [x] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

## What is the current behaviour?

Closes #804

## What is the new behaviour?

Every dispatchEvent of input event now passes the bubbling stage, and it works with React

### DIFF
--- a/projects/core/src/lib/utils/dom/set-element-value.ts
+++ b/projects/core/src/lib/utils/dom/set-element-value.ts
@@ -1,0 +1,28 @@
+/**
+ * Sets value to element, and dispatches input event
+ *
+ * @example
+ * maskitoSetElementValue(input, newValue);
+ *
+ * @see {@link https://github.com/taiga-family/maskito/issues/804 issue}
+ *
+ * @return void
+ */
+export function maskitoSetElementValue(
+    element: HTMLInputElement | HTMLTextAreaElement,
+    value: string,
+): void {
+    element.value = value;
+    element.dispatchEvent(
+        new Event(
+            'input',
+            /**
+             * React handles this event only on bubbling phase
+             *
+             * here is the list of events that are processed in the capture stage, others are processed in the bubbling stage
+             * https://github.com/facebook/react/blob/cb2439624f43c510007f65aea5c50a8bb97917e4/packages/react-dom-bindings/src/events/DOMPluginEventSystem.js#L222
+             */
+            {bubbles: true},
+        ),
+    );
+}
--- a/projects/core/src/lib/utils/dom/update-element.ts
+++ b/projects/core/src/lib/utils/dom/update-element.ts
@@ -1,18 +1,30 @@
+import {ElementState} from '../../types';
+
 /**
  * Sets value to element, and dispatches input event
+ * if you passed ELementState, it also sets selection range
  *
  * @example
- * maskitoSetElementValue(input, newValue);
+ * maskitoUpdateElement(input, newValue);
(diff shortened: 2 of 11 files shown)


projects/core/src/lib/utils/dom/set-element-value.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Sets value to element, and dispatches input event
+ *
+ * @example
+ * maskitoSetElementValue(input, newValue);
+ *
+ * @see {@link https://github.com/taiga-family/maskito/issues/804 issue}
+ *
+ * @return void
+ */
+export function maskitoSetElementValue(
```
nsbarsukov, line 11: * https://github.com/taiga-family/maskito/issues/804#issuecomment-1864095197

nsbarsukov: @aktanoff Could you please update this function according to this discussion?

I apologize for the fuss!

projects/core/src/lib/utils/dom/update-element.ts
```diff
@@ -1,18 +1,30 @@
...
  *
  * @see {@link https://github.com/taiga-family/maskito/issues/804 issue}
  *
  * @return void
  */
-export function maskitoSetElementValue(
+export function maskitoUpdateElement(
     element: HTMLInputElement | HTMLTextAreaElement,
-    value: string,
+    valueOrElementState: ElementState | string,
```
splincode, line 17: `state: ElementState | string` is okey I think

nsbarsukov (approved):

PR #937, opened 2024-01-22
fix(core): don't execute `setSelectionRange` if element is not focused

### LINKED ISSUE #929: 🐞 - No need to execute `setSelectionRange` if element is not focused

@maskito/core, @maskito/kit, @maskito/react

### Playground Link

https://stackblitz.com/edit/stackblitz-starters-4s7lyp?file=src%2FNumberInput.tsx

### Description

When I add a postfix in maskitoNumberOptionsGenerator (but I think it also appends if I add a postfix manually through postProcess), the input gains focus without any action.

I have only seen this in the mobile version because the keyboard opens automatically. You can try it using the StackBlitz link; go to the application link on a mobile device.

### Maskito version

1.9.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #929


projects/core/src/lib/utils/dom/get-focused.ts
```diff
@@ -3,7 +3,7 @@
  *
  * @return element or null
  */
-export function getFocused({activeElement}: Document): Element | null {
+export function maskitoGetFocused({activeElement}: Document): Element | null {
```
waterplea, line 6: Why do we expose it now?

waterplea (approved):

projects/core/src/lib/utils/dom/get-focused.ts
nsbarsukov: I need this utility in both `@maskito/core` & `@maskito/kit` packages (

nsbarsukov: @waterplea review again, please

waterplea (approved):

PR #986, opened 2024-02-01
fix(kit): `Number` with initial value has problems with first time input

nsbarsukov:
# Before
See new Cypress tests – they both failed before fix.


projects/demo-integrations/src/tests/component-testing/number/with-initial-value.cy.ts
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    it('$6 432 kg => Select all + Backspace => $| kg', () => {
+        cy.mount(TestInput, {
+            componentProperties: {maskitoOptions, initialValue: '$6 432 kg'},
+        });
+
+        cy.get('input')
+            .type('{selectAll}{backspace}')
+            .should('have.value', '$ kg')
```
nsbarsukov, line 38: Without fix

first-fail

projects/demo-integrations/src/tests/component-testing/number/with-initial-value.cy.ts
```diff
@@ -0,0 +1,61 @@
...
+        cy.get('input')
+            .focus()
+            .type('{moveToEnd}')
+            .should('have.prop', 'selectionStart', '$6 432'.length)
+            .should('have.prop', 'selectionEnd', '$6 432'.length)
+            .type('{leftArrow}'.repeat(2))
+            .should('have.prop', 'selectionStart', '$6 4'.length)
+            .should('have.prop', 'selectionEnd', '$6 4'.length)
+            .type('{del}')
+            .should('have.value', '$642 kg')
```
nsbarsukov, line 57: Without fix
second-fail

splincode (approved):

PR #1083, opened 2024-02-24
deprecate(angular): `MaskitoCVA` is no longer needed (just use `MaskitoDirective`)

demensky:
Now the `MaskitoDirective` directive can perform the functions of the `MaskitoCVA` directive.

This change will allow us to avoid the situation where the developer forgot to import `MaskitoCVA`, now it just works using `MaskitoDirective`.

`MaskitoCVA` was left as a placeholder to avoid a breaking change.


projects/angular/src/lib/maskito.directive.ts
```diff
@@ -6,29 +6,49 @@ import {
...
     constructor(
         @Inject(NgZone) private readonly ngZone: NgZone,
         @Inject(ElementRef) private readonly elementRef: ElementRef<HTMLElement>,
-    ) {}
+        @Inject(DefaultValueAccessor)
+        @Self()
+        @Optional()
+        accessor: DefaultValueAccessor | null,
+    ) {
+        if (accessor !== null) {
```
waterplea, line 40: nit: we can save a byte or two by writing just `if (accessor)` :)

waterplea (approved):

nsbarsukov (approved):

projects/angular/src/lib/maskito.directive.ts
nsbarsukov:
```suggestion
        if (accessor) {
```

Issue #886, opened 2024-01-10
🤖 - Rewrite some Cypress tests using component testing


nsbarsukov:
### Which package(s) are relevant/related to the feature request?

Don't known / other

### Description

https://github.com/taiga-family/maskito/blob/3edfd30d93def042fb8c28770e9b75e8bd086df0/projects/demo/src/app/app.routes.ts#L276-L283

**Algorithm:**
1. Take any folder from [`projects/demo/src/pages/cypress/examples`](https://github.com/taiga-family/maskito/tree/main/projects/demo/src/pages/cypress/examples)
2. Crete folder with the same name (but without digit) inside [`projects/demo-integrations/src/tests/component-testing`](https://github.com/taiga-family/maskito/tree/main/projects/demo-integrations/src/tests/component-testing).
3. Rewrite all tests (connected with this example) using Cypress component testing (use `DemoPath.Cypress` to find these tests)
4. Delete example from step 1
5. Open PR and link it to this issue

PR #1100, opened 2024-02-28
refactor(demo-integrations): rewriting e2e tests to CT (tests for `predicate` Maskito attribute)

KrollikRoddzer:
Partially solves https://github.com/taiga-family/maskito/issues/886
Done:
1. Added maskitoElement attribute and maskitoElementPredicate property for TestInput
2. Rewrote tests for predicate from e2e to CT


projects/demo-integrations/src/tests/component-testing/predicate/predicate.cy.ts
```diff
@@ -0,0 +1,66 @@
...
+    };
+
+    const nameMask: MaskitoOptions = {
+        mask: /^[a-zA-Z\s]+$/,
+        postprocessors: [
+            ({value, selection}) => ({value: value.toUpperCase(), selection}),
+        ],
+    };
+
+    describe('can detect run-time changes', () => {
```
nsbarsukov, line 25: Unfortunately, the new version of tests (inside this spec) are not equivalent to previous ones.

For these tests you should create the separate test component with the following template:
https://github.com/taiga-family/maskito/blob/5ac91a5d648bbce753cca0568b51ed66dbc81517/projects/demo/src/pages/cypress/examples/1-predicate/template.html#L1-L13

nsbarsukov (changes requested):

projects/demo-integrations/src/tests/component-testing/predicate/predicate.cy.ts
KrollikRoddzer: done

projects/demo-integrations/src/tests/component-testing/predicate/synchronous-test-input.component.ts
```diff
@@ -0,0 +1,54 @@
...
+            <input
+                #card
+                [(ngModel)]="value.number"
+            />
+            <input [(ngModel)]="value.name" />
+        </div>
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SynchronousTestInputComponent {
```
nsbarsukov, line 24: Why `Synchronous`?

May be something like `MultiTestInput`?

projects/demo-integrations/src/tests/component-testing/predicate/predicate.cy.ts
```diff
@@ -0,0 +1,46 @@
+import {MaskitoOptions} from '@maskito/core';
+
+import {TestInput} from '../utils';
+import {SynchronousTestInputComponent} from './synchronous-test-input.component';
+
+describe('@maskito/angular | Predicate', () => {
+    const cardMask: MaskitoOptions = {
```
nsbarsukov, line 7: Is it used only inside `it('supports asynchronous predicate', ...)` ?

Let's move it inside this spec.

projects/demo-integrations/src/tests/component-testing/predicate/synchronous-test-input.component.ts
KrollikRoddzer: Because the other test in asynchronous. I renamed it.

projects/demo-integrations/src/tests/component-testing/predicate/predicate.cy.ts
KrollikRoddzer: Moved

nsbarsukov (approved):

PR #1167, opened 2024-03-20
chore: fix IDE error when import `is not under rootDir`

nsbarsukov:
# Previous behavior
before

# New behavior
after

### DIFF
--- a/projects/core/tsconfig.lib.json
+++ b/projects/core/tsconfig.lib.json
@@ -1,6 +1,3 @@
 {
-    "extends": "../../tsconfig.build.json",
-    "compilerOptions": {
-        "rootDir": "."
-    }
+    "extends": "../../tsconfig.build.json"
 }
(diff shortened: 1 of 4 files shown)


projects/core/tsconfig.lib.json
```diff
@@ -1,6 +1,3 @@
 {
-    "extends": "../../tsconfig.build.json",
-    "compilerOptions": {
-        "rootDir": "."
-    }
+    "extends": "../../tsconfig.build.json"
 }
```
nsbarsukov, line 3: If we:

1. Delete this file totally 
2. Replace this line
https://github.com/taiga-family/maskito/blob/efa06301c78c4d387c1a3826ae76c3a1887730f5/projects/core/project.json#L14
with the following one:
```json
 "tsConfig": "tsconfig.build.json", 
```

`nx build core` will throw error:

```
Error during bundle:
Error: error TS6059:
File '/projects/angular/src/test-setup.ts' is not under 'rootDir' '/projects/core'.
'rootDir' is expected to contain all source files.
  The file is in the program because:
    Root file specified for compilation
error TS6059: File '/projects/demo-integrations/cypress.config.ts' is not under 'rootDir' '/maskito/projects/core'. 'rootDir' is expected to contain all source files.
# [...]
```

splincode (approved):

PR #1337, opened 2024-06-25
chore: reproduction for Nx bug (it installs deps with `legacy-peer-deps=true`)

nsbarsukov:
## Getting started
```
nvm use 18
```

```
npm ci
```

Pay attention that project does **NOT** use [legacy-peer-deps](https://docs.npmjs.com/cli/v10/using-npm/config#legacy-peer-deps)!

## Case 1
Run the following command
```
nx migrate --run-migrations
```

it will logs:
```
 NX   Running 'npm install' to make sure necessary packages are installed

removed 297 packages, and audited 2714 packages in 4s
```

and then it will throw error:
```
Cannot find module 'ts-morph'
```

## Case 2
Reset all environment (revert `package-lock.json` changes and run again `npm ci`).

Then run the following command:
```
npm_config_legacy_peer_deps=false nx migrate --run-migrations
```

No error. Migrations are executed successfully!


nsbarsukov:
DONT MERGE

migrations.json
```diff
@@ -0,0 +1,25 @@
+{
```
splincode, line 1: add to `.gitignore`

PR #1338, opened 2024-06-25
feat(core): new built-in `maskitoChangeEventPlugin`

### LINKED ISSUE #1329: 🐞 - New built-in plugin `maskitoChangeEventPlugin` (to dispatch native `change` event)

@maskito/core, @maskito/kit

### Playground Link

https://um2sal.stackblitz.io/

### Description

I need to enter at least 4 digits + decimal seperator in order to have the onchange function triggered.

If you enter for example
",1"
",12"
"0,12"
"1,22"

The onchange function is not triggered

Only if I enter

"12,34"
"12," => which gets transformed into "12,00"
"01,00"
"01," => which gets transformed into "1,00"

the on change function is triggered. Also deleting every digit will trigger the onchange.

### Maskito version

latest

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
Closes #1329

### DIFF
--- a/projects/core/src/lib/plugins/change-event-plugin.ts
+++ b/projects/core/src/lib/plugins/change-event-plugin.ts
@@ -8,25 +8,22 @@ export function maskitoChangeEventPlugin(): MaskitoPlugin {
 
         let value = element.value;
 
-        const focusListener = (): void => {
-            value = element.value;
-        };
-        const changeListener = (): void => {
+        const valueListener = (): void => {
             value = element.value;
         };
         const blurListener = (): void => {
             if (element.value !== value) {
-                element.dispatchEvent(new Event('change'));
+                element.dispatchEvent(new Event('change', {bubbles: true}));
             }
         };
 
-        element.addEventListener('focus', focusListener);
-        element.addEventListener('change', changeListener);
+        element.addEventListener('focus', valueListener);
+        element.addEventListener('change', valueListener);
         element.addEventListener('blur', blurListener);
 
         return () => {
-            element.removeEventListener('focus', focusListener);
-            element.removeEventListener('change', changeListener);
+            element.removeEventListener('focus', valueListener);
+            element.removeEventListener('change', valueListener);
             element.removeEventListener('blur', blurListener);
         };
     };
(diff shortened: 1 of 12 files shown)


projects/core/src/lib/plugins/change-event-plugin.ts
```diff
@@ -0,0 +1,33 @@
...
+
+        const focusListener = (): void => {
+            value = element.value;
+        };
+        const changeListener = (): void => {
+            value = element.value;
+        };
+        const blurListener = (): void => {
+            if (element.value !== value) {
+                element.dispatchEvent(new Event('change'));
```
waterplea, line 19: Change event bubbles, need to add `{ bubbles: true }`

projects/core/src/lib/plugins/change-event-plugin.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function maskitoChangeEventPlugin(): MaskitoPlugin {
+    return element => {
+        if (element.isContentEditable) {
+            return;
+        }
+
+        let value = element.value;
+
+        const focusListener = (): void => {
```
waterplea, line 11: Let's call it `valueListener`  and use the same on both cases, since there's no difference?

projects/core/src/lib/plugins/change-event-plugin.ts
nsbarsukov: Fixed in https://github.com/taiga-family/maskito/pull/1338/commits/55192a78648be4d4c17068c53347d8ca5af5115a

projects/core/src/lib/plugins/change-event-plugin.ts
nsbarsukov: Fixed in https://github.com/taiga-family/maskito/pull/1338/commits/55192a78648be4d4c17068c53347d8ca5af5115a

waterplea (approved):

PR #1398, opened 2024-07-19
chore(demo): using standalone API to bootstrap demo application

nsbarsukov:


projects/demo/server.ts
```diff
@@ -6,75 +6,45 @@ import {join} from 'node:path';
...
-        : 'index';
-
-    // Our Universal express-engine (found @ https://github.com/angular/universal/tree/master/modules/express-engine)
-    server.engine(
-        'html',
-        ngExpressEngine({
-            bootstrap: AppServerModule,
-        }) as any,
-    );
+declare const __non_webpack_require__: NodeRequire;
```
nsbarsukov, line 13: All content from this file is copy-pasted from Taiga repo:
https://github.com/taiga-family/taiga-ui/blob/main/projects/demo/server.ts

splincode (approved):

PR #1401, opened 2024-07-22
fix(kit): `maskitoStringifyTime` was adding `0` on the wrong side

### LINKED ISSUE #1399: 🐞 - maskitoStringifyTime convert to wrong time

@maskito/kit

### Playground Link

https://stackblitz.com/edit/79evv4?file=index.ts (see console.log)

### Description

When you convert from "parse time" to "stringify time" some combinations are put wrong, for example:
03:02:00 generates 10920000
But when you try to convert to string it returns:
30:20:00 instead of 03:02:00.

I believe the problem maybe is in the padTimeSegments, which should put "0" at the beginning and not at the end, but messing with this breaks all parse-time.ts tests.

### Maskito version

2.5.0 and 3.0.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

demensky:
Closes #1399

### DIFF
--- a/projects/kit/src/lib/masks/date-time/preprocessors/valid-date-time-preprocessor.ts
+++ b/projects/kit/src/lib/masks/date-time/preprocessors/valid-date-time-preprocessor.ts
@@ -3,7 +3,7 @@ import type {MaskitoPreprocessor} from '@maskito/core';
 import {DEFAULT_TIME_SEGMENT_MAX_VALUES, TIME_FIXED_CHARACTERS} from '../../../constants';
 import type {MaskitoTimeMode} from '../../../types';
 import {escapeRegExp, validateDateString} from '../../../utils';
-import {padTimeSegments, validateTimeString} from '../../../utils/time';
+import {padEndTimeSegments, validateTimeString} from '../../../utils/time';
 import {parseDateTimeString} from '../utils';
 
 export function createValidDateTimePreprocessor({
@@ -66,7 +66,7 @@ export function createValidDateTimePreprocessor({
 
         validatedValue += validatedDateString;
 
-        const paddedMaxValues = padTimeSegments(DEFAULT_TIME_SEGMENT_MAX_VALUES);
+        const paddedMaxValues = padEndTimeSegments(DEFAULT_TIME_SEGMENT_MAX_VALUES);
 
         const {validatedTimeString, updatedTimeSelection} = validateTimeString({
             timeString,
--- a/projects/kit/src/lib/masks/date-time/preprocessors/valid-date-time-preprocessor.ts
+++ b/projects/kit/src/lib/masks/date-time/preprocessors/valid-date-time-preprocessor.ts
@@ -3,7 +3,7 @@ import type {MaskitoPreprocessor} from '@maskito/core';
 import {DEFAULT_TIME_SEGMENT_MAX_VALUES, TIME_FIXED_CHARACTERS} from '../../../constants';
 import type {MaskitoTimeMode} from '../../../types';
 import {escapeRegExp, validateDateString} from '../../../utils';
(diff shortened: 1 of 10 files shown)


projects/kit/src/lib/masks/time/processors/max-validation-preprocessor.ts
```diff
@@ -3,13 +3,13 @@ import type {MaskitoPreprocessor} from '@maskito/core';
...
 import {escapeRegExp} from '../../../utils';
-import {padTimeSegments, validateTimeString} from '../../../utils/time';
+import {padEndTimeSegments, validateTimeString} from '../../../utils/time';
 
 export function createMaxValidationPreprocessor(
     timeSegmentMaxValues: MaskitoTimeSegments<number>,
     timeMode: MaskitoTimeMode,
 ): MaskitoPreprocessor {
-    const paddedMaxValues = padTimeSegments(timeSegmentMaxValues);
+    const paddedMaxValues = padEndTimeSegments(timeSegmentMaxValues);
```
nsbarsukov, line 12: If I've not mistaken, `padStartTimeSegments` is more suitable here 🤔 

Probably, case with single digit was always broken :D

https://stackblitz.com/edit/time-segment-max-value-bug-single-digit?file=mask.ts

```ts
const timeOptions = maskitoTimeOptionsGenerator({
    mode: 'HH:MM',
    timeSegmentMaxValues: {hours: 4}, // you can enter 09:09 :(
});
```

https://stackblitz.com/edit/time-segment-max-value-bug-single-digit?file=mask.ts

nsbarsukov:
```suggestion
    const paddedMaxValues = padStartTimeSegments(timeSegmentMaxValues);
```

\+ add some cypress tests for single-digit cases, please

projects/kit/src/lib/masks/date-time/preprocessors/valid-date-time-preprocessor.ts
```diff
@@ -66,7 +66,7 @@ export function createValidDateTimePreprocessor({
 
         validatedValue += validatedDateString;
 
-        const paddedMaxValues = padTimeSegments(DEFAULT_TIME_SEGMENT_MAX_VALUES);
+        const paddedMaxValues = padEndTimeSegments(DEFAULT_TIME_SEGMENT_MAX_VALUES);
```
nsbarsukov, line 69:
```suggestion
        const paddedMaxValues = padStartTimeSegments(DEFAULT_TIME_SEGMENT_MAX_VALUES);
```

nsbarsukov (changes requested):

projects/kit/src/lib/masks/date-time/preprocessors/valid-date-time-preprocessor.ts
demensky: This change does not affect anything, but it will be useful for code clarity.

projects/kit/src/lib/masks/time/processors/max-validation-preprocessor.ts
demensky: Updated. Unfortunately, Cypress is no longer with me.

nsbarsukov: > Updated. Unfortunately, Cypress is no longer with me.

I'll do it later in separate PR 👌

nsbarsukov (approved):

nsbarsukov:
@demensky thanks ❤️

PR #1454, opened 2024-08-15
chore(demo): update `Taiga UI` to stable version of `v4`

nsbarsukov:


projects/demo/src/styles.less
```diff
@@ -6,7 +6,7 @@ body {
 
 html,
 body {
-    .scroll-behavior();
+    scroll-behavior: var(--tui-scroll-behavior);
```
splincode, line 9: `scroll-behavior: smooth;`

splincode (approved):

PR #1608, opened 2024-09-13
fix(angular): race condition when `[maskitoOptions]` are changed before long element predicate is resolved 

TheHomeGera:
fix dynamic options with customplugins maskitoEventHandler.

When creating new options without destroying the control, subscriptions to the element multiply. When creating new options, you must first unsubscribe from them.


splincode (approved):

projects/angular/src/lib/maskito.directive.ts
```diff
@@ -55,6 +55,7 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
         }
 
         ngZone.runOutsideAngular(() => {
+            this.maskedElement?.destroy();
```
splincode, line 58: @TheHomeGera can you add unit test for catch this?

TheHomeGera: sorry, I can't to write unit test on it, maybe cypress can catch it. 

The picture shows an example of how to catch. If the second options flies into ngOnChanges before the callback from runOutsideAngular is executed. Then destroy will not work, since destroy has worked, and there will be a double field setting already in the callback.
image

splincode: `maybe cypress can catch it.` could you add?

nsbarsukov:
👀

projects/angular/src/lib/maskito.directive.ts
nsbarsukov, line 58: 👀

nsbarsukov, line 58: Let's revert this line and extend this condition instead 
https://github.com/taiga-family/maskito/blob/691b4eb84eeaee45f4b9e0be834e1100e1c0834b/projects/angular/src/lib/maskito.directive.ts#L51-L55

⬇️ 
```ts
if (
  this.elementPredicate !== elementPredicate ||
  this.options !== options
) {
  // maskito element (or its options) has changed before the predicate was resolved.
  return;
}
```

nsbarsukov (changes requested):
Thanks for your contribution ❤️

projects/angular/src/lib/maskito.directive.ts
nsbarsukov: @TheHomeGera Hello!

I've written some Cypress tests for this case:
* https://github.com/taiga-family/maskito/pull/1630

Just rebase your branch and enable them in your PR.

You need to remove `.skip` (and `TODO` comments) for these tests:
https://github.com/taiga-family/maskito/blob/92a8f3208288e01648757c7c2e5f56867d404100/projects/demo-integrations/src/tests/component-testing/angular-predicate/angular-predicate.cy.ts#L178-L179

https://github.com/taiga-family/maskito/blob/92a8f3208288e01648757c7c2e5f56867d404100/projects/demo-integrations/src/tests/component-testing/angular-predicate/angular-predicate.cy.ts#L184-L185

TheHomeGera: +

nsbarsukov: @TheHomeGera Hello!
Could you update your PR according to this thread ?

nsbarsukov:
Closing this issue in favor of this PR:
* https://github.com/taiga-family/maskito/pull/1696

PR #1796, opened 2024-10-31
fix(kit): `DateRange` should accept single character date segment paste even if date and range separators are equal

### LINKED ISSUE #1256: 🐞 - `DateRange` should accept single character date segment during paste when date separated by '-' sign

@maskito/kit

### Playground Link

https://maskito.dev/kit/date-range/API

### Description

Original issue https://github.com/taiga-family/maskito/issues/593 fixes the problem for dates formatted with dots like `11.9.2022`.
The bug still can be reproduced with dates like `11-9-2022` or `7-10-2020` for `DateRange` only. Works fine for `Date` and `DateTime`.

Steps to reproduce
https://maskito.dev/kit/date-range/API?mode=dd%2Fmm%2Fyyyy

Paste 11-9-2022
✅ Expected behaviour:
Textfield contains 11.09.2022

❌ Actual behaviour:
Textfield contains 11.09.022

### Maskito version

2.3.2

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

AndreiBelokopytov:
Fixes #1256


projects/kit/src/lib/processors/normalize-date-preprocessor.ts
```diff
@@ -14,53 +14,52 @@ export function normalizeDatePreprocessor({
...
-            };
+        for (let index = 0; index < dateSegments.length; index++) {
+            const segment = dateSegments[index]!;
+            const template = templateSegments[index % templateSegments.length];
+            const isLastTemplateSegment =
+                index % templateSegments.length === templateSegments.length - 1;
+
+            if (index >= templateSegments.length && includesTime) {
+                timeParts.push(segment);
+                continue;
```
nsbarsukov, line 34: I have an idea how to make code more readable with less variable mutations.

```ts
const dateParts: string[] = segments
    .slice(0, includesTime ? templateSegments.length : Infinity)
    .map((segment) => {
        // ...
    });
const timeParts: string[] = segments
    .slice(includesTime ? templateSegments.length : Infinity)
    .map((segment) => {
        // ...
    });
```

What do you think ? 🤔

projects/kit/src/lib/processors/normalize-date-preprocessor.ts
```diff
@@ -14,53 +14,52 @@ export function normalizeDatePreprocessor({
...
-                    date.trim().split(/\D/).filter(Boolean).length ===
-                    dateModeTemplate.split(dateSegmentsSeparator).length,
-            )
-        ) {
-            const newData = dates
-                .map((date) =>
-                    normalizeDateString(date, dateModeTemplate, dateSegmentsSeparator),
-                )
-                .join(rangeSeparator);
+        let newData = data;
```
nsbarsukov, line 21: Do we actually need to create it here?

Could we just write in following way ?
```ts
if (firstCodition) {
    return {elementState, date: anotherCondition ? 'option1' : 'option2'};
}

return {elementState, data};
```

projects/kit/src/lib/processors/normalize-date-preprocessor.ts
```diff
@@ -14,53 +14,52 @@ export function normalizeDatePreprocessor({
...
 }): MaskitoPreprocessor {
     return ({elementState, data}) => {
-        const separator = rangeSeparator
-            ? new RegExp(`${rangeSeparator}|-`)
-            : dateTimeSeparator;
-        const possibleDates = data.split(separator);
-        const dates = data.includes(dateTimeSeparator)
-            ? [possibleDates[0] ?? '']
-            : possibleDates;
+        const dateSegments = data.split(/\D/).filter(Boolean);
```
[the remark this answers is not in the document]
AndreiBelokopytov: Hmm... I'am concerned `segements` might be confused with `segment`.
So may be `dateTimeSegments`?

projects/kit/src/lib/processors/normalize-date-preprocessor.ts
```diff
@@ -14,53 +14,45 @@ export function normalizeDatePreprocessor({
...
-            : segment.padStart(templateSegments[index]?.length ?? 0, '0'),
-    );
+            if (!dates[dateIndex]) {
+                dates[dateIndex] = '';
+            }
 
-    return normalizedSegments.join(separator);
+            dates[dateIndex] += isLastDateSegment
+                ? segment
+                : `${segment.padStart(template?.length ?? 0, '0')}${dateSegmentsSeparator}`;
```
nsbarsukov, line 46:
```ts
const template = templateSegments[index % templateSegments.length] ?? '';
// ...
`${segment.padStart(template.length, '0')}${dateSegmentsSeparator}`;
```

projects/kit/src/lib/processors/normalize-date-preprocessor.ts
```diff
@@ -14,53 +14,45 @@ export function normalizeDatePreprocessor({
...
+
+            return dates;
+        }, []);
+
+        return {
+            elementState,
+            data: includesTime
+                ? `${dates[0]}${dateTimeSeparator}${timeSegments.join(':')}`
```
nsbarsukov, line 54: It is not true for all time modes – time mode can contain dot too.

https://github.com/taiga-family/maskito/blob/20fce011c67173ab3155a7145ae4299405943917/projects/kit/src/lib/types/time-mode.ts#L10

nsbarsukov:
```suggestion
                ? `${dates[0]}${data.slice(dateTimeSeparator)}`
```

\+ delete no more required `timeSegments` array

nsbarsukov: Add some unit tests for `DateTime` (with `MM:SS.MSS` mode) inside `normalize-date-preprocessor.spec.ts`, please

nsbarsukov: It also means that we can delete `dateTimeSegments` variable

```ts
const dateSegments = data
    .split(/\D/)
    .filter(Boolean)
    .slice(0, includesTime ? templateSegments.length : Infinity);
```

nsbarsukov (approved):
Thanks ❤️

PR #1865, opened 2024-11-25
fix: prevent extra spaces from being added by postfix-postprocessor

### LINKED ISSUE #1820: 🐞 - `Number` + postfix (with leading space) adds unnecessary spaces on paste of decimal number

@maskito/kit

### Playground Link

https://stackblitz.com/edit/bsm8fn?file=index.ts&hideExplorer=1

### Description

When you use these settings to configure Maskito:

```typescript
const maskitoOptions: MaskitoOptions = maskitoNumberOptionsGenerator({
  postfix: ' EUR',
  thousandSeparator: '.',
  decimalSeparator: ',',
  precision: 2,
  decimalZeroPadding: true,
  minusSign: '-',
});
```
Then its possible to paste extra spaces into the field between the postfix and the number.
This only happens when the field is completely empty.

# Steps:
- Load Page and see empty maskito input
  Image
- paste `1234,12   ` into the input (with trailing spaces)

**Expected:**
Image

**Saw:**
Image

### Maskito version

3.2.0

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [x] Edge

- [ ] macOS
- [x] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

AndreiBelokopytov:
Fixes #1820

The addition of an extra space happens in the postfix postprocessor because it believes the mask was changed and tries to find a symbol in the original string that also exists in the mask.  

To fix this behavior, an additional check has been added to ensure that the mask wasn't changed: the value should not contain the mask.

UPD: removed several unit tests due to limitations in the `postfixPostprocessor`.  Currently, it lacks the information needed to preserve deleted symbols that are not part of the mask.  

While `extractAffixes` now correctly handles overlapping prefixes and suffixes, modifying postfixPostprocessor's signature to accept a prefix would be a more robust solution.


projects/core/src/lib/classes/mask-model/utils/guess-valid-value-by-reg-exp.ts
```diff
@@ -22,5 +22,13 @@ export function guessValidValueByRegExp(
...
 
+    if (newFrom > validatedValue.length) {
+        newFrom = validatedValue.length;
+    }
+
+    if (newTo > validatedValue.length) {
+        newTo = validatedValue.length;
+    }
+
     return {value: validatedValue, selection: [newFrom, newTo]};
```
nsbarsukov, line 33:
```ts
return {
    value: validatedValue,
    selection: [
        Math.min(newFrom, validatedValue.length),
        Math.min(newTo, validatedValue.length),
    ],
};
```

projects/kit/src/lib/utils/extract-affixes.ts
```diff
@@ -14,7 +14,16 @@ export function extractAffixes(
...
+            extractedPostfix,
+            cleanValue: value.slice(
+                extractedPrefix.length,
+                extractedPostfix.length ? -extractedPostfix.length : Infinity,
+            ),
+        };
+    }
 
-    return {extractedPrefix, extractedPostfix, cleanValue};
+    return {extractedPrefix, extractedPostfix, cleanValue: value};
```
nsbarsukov, line 28:
```ts
return {
    extractedPrefix,
    extractedPostfix,
    cleanValue:
        extractedPrefix || extractedPostfix
            ? value.slice(
                  extractedPrefix.length,
                  extractedPostfix.length ? -extractedPostfix.length : Infinity,
              )
            : value,
};
```

projects/demo-integrations/src/tests/component-testing/number/mirrored-value-postfix.cy.ts
```diff
@@ -0,0 +1,34 @@
...
+                maskitoOptions: maskitoNumberOptionsGenerator({
+                    postfix: ' EUR',
+                    precision: 2,
+                }),
+            },
+        });
+        cy.get('input').focus().should('have.value', '').as('input');
+    });
+
+    it('Empty input => Paste "11.22 " => Textfield\'s value is "11.22 EUR"', () => {
```
nsbarsukov, line 19:
```suggestion
    it('Empty input => Paste "11.22 " => 11.22 |EUR', () => {
```

projects/demo-integrations/src/tests/component-testing/number/mirrored-value-postfix.cy.ts
```diff
@@ -0,0 +1,34 @@
...
+        cy.get('input')
+            .paste('11.22 ')
+            .should('have.value', '11.22 EUR')
+            .should('have.prop', 'selectionStart', '11.22 '.length)
+            .should('have.prop', 'selectionEnd', '11.22 '.length);
+    });
+
+    it('Empty input => Paste "11.22  " => Textfield\'s value is "11.22 EUR"', () => {
```
nsbarsukov, line 27:
```suggestion
    it('Empty input => Paste "11.22  " (with two trailing spaces) => 11.22 |EUR', () => {
```

PR #1939, opened 2025-01-28
feat(kit): `Date` supports `dd/mm` and `mm/dd` modes

### LINKED ISSUE #1827: 🚀 - `MaskitoDateMode` supports `dd/mm` & `mm/dd` formats

@maskito/kit

### Description

`maskitoDateOptionsGenerator` function do not allow dd/mm and mm/dd date formats.

Because function `createMinMaxDatePostprocessor` has default `min` date set to `new Date('0001-01-01')`
If user don't input year then `clamp` function set the date to `min` date and return `12/31`.

To allow this, we can remove the `DEFAULT_MIN_DATE` from `min` parameter in function `createMinMaxDatePostprocessor`.

and then return a date if `min` is `undefined`
`const clampedDate = min ? clamp(date, min, max) : date;`
https://github.com/taiga-family/maskito/blob/c81cda5d3c2e6f5f87aedfa984f7f235edca8c2b/projects/kit/src/lib/processors/min-max-date-postprocessor.ts#L52

AndreiBelokopytov:
Fixes #1827


projects/kit/src/lib/processors/min-max-date-postprocessor.ts
```diff
@@ -48,7 +50,12 @@ export function createMinMaxDatePostprocessor({
                 continue;
             }
 
+            if (!parsedDate.year) {
+                parsedDate.year = LEAP_YEAR;
+            }
+
             const date = segmentsToDate(parsedDate);
```
nsbarsukov, line 57:
```suggestion
            const date = segmentsToDate({year: LEAP_YEAR, ...parsedDate});
```

nsbarsukov (approved):

PR #1986, opened 2025-03-02
chore: update `keywords` for publishable libraries

ggomarighetti:
### Resume
I was looking for an updated library to perform input masking in react using the proposed keywords. I could only find this library after having reached “text-mask” through reddit and seeing it as an option in the list of alternatives.

It seems to me that by updating the list of keywords (I don't know if there are restrictions on the number or content) the library could have more visibility and more contributors, which it deserves since it managed to solve my problems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
     "version": "3.3.0",
     "description": "Collection of libraries to create an input mask which ensures that user types value according to predefined format",
     "keywords": [
+        "input",
         "mask",
         "inputmask",
         "input-mask",
@@ -10,7 +11,8 @@
         "input-formatting",
         "javascript",
         "typescript",
-        "angular"
+        "angular",
+        "react"
     ],
     "homepage": "https://maskito.dev",
     "bugs": "https://github.com/taiga-family/maskito/issues",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
     "name": "maskito",
     "version": "3.3.0",
     "description": "Collection of libraries to create an input mask which ensures that user types value according to predefined format",
-    "keywords": [],
     "homepage": "https://maskito.dev",
     "bugs": "https://github.com/taiga-family/maskito/issues",
     "repository": {
--- a/package.json
+++ b/package.json
@@ -2,18 +2,7 @@
     "name": "maskito",
     "version": "3.3.0",
     "description": "Collection of libraries to create an input mask which ensures that user types value according to predefined format",
-    "keywords": [
-        "input",
-        "mask",
-        "inputmask",
-        "input-mask",
-        "text-mask",
-        "input-formatting",
-        "javascript",
-        "typescript",
-        "angular",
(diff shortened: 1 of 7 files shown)


ggomarighetti:
I modified only the keywords that I used at the time of the search and the core package, I understand that you could modify the existing ones, adding variants or also presenting separately the existing ones, for example:

`input-formatting` —> `input`(already) `formatting` `formatting` `input-format` `input-formatting`(already), etc...

`text-mask` —> `mask` (already) `text-mask` `number-mask` `date-mask` `currency-mask`, etc...

projects/core/package.json
```diff
@@ -3,13 +3,15 @@
     "version": "3.3.0",
     "description": "The main zero-dependency and framework-agnostic Maskito's package to create an input mask",
     "keywords": [
+        "input",
```
nsbarsukov, line 6: Add this keyword for other libraries too

https://github.com/taiga-family/maskito/blob/5640ed3a332b2816d8aad9fe083fb930769d7bef/projects/angular/package.json#L2-L5

https://github.com/taiga-family/maskito/blob/5640ed3a332b2816d8aad9fe083fb930769d7bef/projects/vue/package.json#L2-L5

https://github.com/taiga-family/maskito/blob/5640ed3a332b2816d8aad9fe083fb930769d7bef/projects/kit/package.json#L2-L5

https://github.com/taiga-family/maskito/blob/5640ed3a332b2816d8aad9fe083fb930769d7bef/projects/phone/package.json#L2-L5

projects/core/package.json
```diff
@@ -3,13 +3,15 @@
...
         "mask",
         "inputmask",
         "input-mask",
         "text-mask",
         "input-formatting",
         "javascript",
-        "typescript"
+        "typescript",
+        "react",
```
nsbarsukov, line 14: I am not sure that we should add this keyword for framework-agnostic library `@maskito/core`.

`@maskito/react` already has this keyword – I think it is enough for any user search
https://github.com/taiga-family/maskito/blob/5640ed3a332b2816d8aad9fe083fb930769d7bef/projects/react/package.json#L2-L11

package.json
```diff
@@ -3,14 +3,16 @@
     "version": "3.3.0",
     "description": "Collection of libraries to create an input mask which ensures that user types value according to predefined format",
     "keywords": [
```
nsbarsukov, line 5: Root `package.json` is not used for any publishable library.
It means that property `keywords` does not make sense for it 😄 

Let's remove it totally for this file.

package.json
```diff
@@ -2,16 +2,7 @@
...
-        "mask",
-        "inputmask",
-        "input-mask",
-        "text-mask",
-        "input-formatting",
-        "javascript",
-        "typescript",
-        "angular"
-    ],
+    "keywords": [],
```
nsbarsukov, line 5: Remove this line

```suggestion
```

Issue #2007, opened 2025-03-19
🐞 - `maskitoParseDate` should return `null` for incompleted date string


nsbarsukov:
### Which package(s) are the source of the bug?

@maskito/kit

### Playground Link

https://maskito.dev/kit/date

### Current behavior
```ts
maskitoParseDate('19', {mode: 'dd/mm/yyyy', separator: '.'}) // new Date(1, 0, 1) ❌

maskitoParseDate('19.03.2', {mode: 'dd/mm/yyyy', separator: '.'}) // new Date(2, 2, 19) ❌

maskitoParseDate('19.03.20', {mode: 'dd/mm/yyyy', separator: '.'}) // new Date(2020, 2, 19) ❌

maskitoParseDate('19.03.2025', {mode: 'dd/mm/yyyy', separator: '.'}) // new Date(2025, 2, 19) ✅
```

### Expected behavior
```ts
maskitoParseDate('19', {mode: 'dd/mm/yyyy', separator: '.'}) // null

maskitoParseDate('19.03.2', {mode: 'dd/mm/yyyy', separator: '.'}) // null

maskitoParseDate('19.03.20', {mode: 'dd/mm/yyyy', separator: '.'}) // null
// BUT!
maskitoParseDate('03.20', {mode: 'mm/yy', separator: '.'}) // new Date(2020, 2)

maskitoParseDate('19.03.2025', {mode: 'dd/mm/yyyy', separator: '.'}) // new Date(2025, 2, 19) 
```

### Possible solution
```ts
export function maskitoParseDate(
    value: string,
    {mode, min = DEFAULT_MIN_DATE, max = DEFAULT_MAX_DATE}: MaskitoDateParams,
): Date {
    if (value.length < mode.length) {
        return null;
    }
    // [...]
}
```

> **Don't forget about unit tests!**

### Maskito version

3.4.0

ggomarighetti:
@nsbarsukov sure,

as I commented in the pull request, the function maskitoParseDate had to return a non-nullable Date type data, as opposed to its equivalent for the number type, which could return NaN.

This caused the function to prematurely return wrong dates, for example, following a mode of “dd/mm/yyyy”, when the first two digits of the year were completed the function already returned a premature result since it entered conditionals that correspond to other modes such as “mm/yy”.

With the pull request the return type of the function is modified as well as the expected result, until the function does not receive a string of characters of the same length as the mode of use, it will not try to parse the date.

nsbarsukov:
[...]

PR #2009, opened 2025-03-19
fix(kit): `maskitoParseDate` should return `null` for incompleted date string

ggomarighetti:
### Resume
The function maskitoParseDate had to return a non-nullable Date type data, as opposed to its equivalent for the number type, which could return NaN.

This caused the function to prematurely return wrong dates, for example, following a mode of “dd/mm/yyyy”, when the first two digits of the year were completed the function already returned a premature result since it entered conditionals that correspond to other modes such as “mm/yy”.

With this pull request the return type of the function is modified as well as the expected result, until the function does not receive a string of characters of the same length as the mode of use, it will not try to parse the date.

### Reference
https://github.com/taiga-family/maskito/issues/2007


nsbarsukov (approved):
Thanks ❤️

PR #2022, opened 2025-03-24
feat(kit): `Number` supports new properties `minimumFractionDigits` & `maximumFractionDigits`

### LINKED ISSUE #2003: 🚀 - `Number` supports new properties `minimumFractionDigits` & `maximumFractionDigits`

@maskito/kit

### Description

**Description**
Hi, first of all, thank you for your work on this library! I have a use case where I need to allow a maximum of 10 decimal places (`precision: 10`), but in most cases, users only enter 2 decimal places. Currently, the input shows all 10 decimal places with trailing zeros when decimalZeroPadding is enabled.

**Proposed Solution**
Allow `decimalZeroPadding` to be configured independently from `precision`, similar to how **Intl.NumberFormat** handles `minimumFractionDigits` and `maximumFractionDigits`.

**Example**

```ts
const options = maskitoNumberOptionsGenerator({
  // ...
  decimalZeroPadding: true,
  minimumFractionDigits: 2,
  maximumFractionDigits: 10,
})
```

1. Input is empty
2. User enters `1`
3. Input shows `1|,00` (as `minimumFractionDigits` is 2)
3. User moves cursor to the end of the input: `1,00|`
4. User can enter up to 10 decimal places: `1,0000000000|` (as `maximumFractionDigits` is 10)

nsbarsukov:
Fixes #2003

### DIFF
--- a/projects/demo/src/pages/kit/number/number-mask-doc.template.html
+++ b/projects/demo/src/pages/kit/number/number-mask-doc.template.html
@@ -28,11 +28,11 @@
             id="high-precision"
             heading="High precision"
             [content]="highPrecisionExample1"
-            [description]="precisionDescription"
+            [description]="maximumFractionDigitsDescription"
         >
-            <ng-template #precisionDescription>
+            <ng-template #maximumFractionDigitsDescription>
                 Use
-                <code>precision</code>
+                <code>maximumFractionDigits</code>
                 parameter to configure the number of digits after decimal separator.
             </ng-template>
             <number-mask-doc-example-1 />
@@ -95,8 +95,8 @@
         >
             <ng-template #decimalZeroPaddingDescription>
                 <p class="tui-space_top-0">
-                    Set
-                    <code>decimalZeroPadding: true</code>
+                    Use
+                    <code>minimumFractionDigits</code>
                     to always show trailing zeroes.
                 </p>
                 <p>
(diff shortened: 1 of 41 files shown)


projects/demo/src/pages/kit/number/number-mask-doc.template.html
```diff
@@ -228,41 +204,70 @@
...
+            <ng-template
+                documentationPropertyMode="input"
+                documentationPropertyName="maximumFractionDigits"
+                documentationPropertyType="number"
+                [documentationPropertyValues]="maximumFractionDigitsOptions"
+                [(documentationPropertyValue)]="maximumFractionDigits"
+                (documentationPropertyValueChange)="updateOptions()"
+            >
+                <p>
+                    The maximum number of of digits after
```
nsbarsukov, line 253:
```suggestion
                    The maximum number of digits after
```

projects/demo/src/pages/kit/number/number-mask-doc.template.html
```diff
@@ -228,41 +204,70 @@
...
+                    <code>0</code>
+                    .
                 </p>
             </ng-template>
+
+            <ng-template
+                documentationPropertyMode="input"
+                documentationPropertyName="maximumFractionDigits"
+                documentationPropertyType="number"
+                [documentationPropertyValues]="maximumFractionDigitsOptions"
```
vladimirpotekhin, line 248: nit: Shouldn't there also be a number input here, not a select? It's a little inconsistent right now 
Снимок экрана 2025-04-02 в 11 02 55

vladimirpotekhin (approved):

projects/demo/src/pages/kit/number/number-mask-doc.template.html
nsbarsukov: `[maximumFractionDigits]` is select (not `[tuiInputNumber]`) only because of possible `Infinity`-value

PR #2040, opened 2025-04-04
fix(core): double space bar removes characters

### LINKED ISSUE #2023: 🐞 - Double space bar removes a character on Android

@maskito/core

### Playground Link

https://ionicframework.com/docs/api/input#input-masking

### Description

**Current behavior:** If you press the space bar twice on Android and only numbers and letters are allowed, the last character is removed.

https://github.com/user-attachments/assets/ec8c039f-1e43-42c6-8517-7dbdc4fd89c5

**Expected behavior**: I expect that nothing happens and spaces are simply ignored if they are not allowed.

### Maskito version

3.4.0

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [x] Android

nsbarsukov:
Fixes #2023

### DIFF
--- a/projects/core/src/lib/plugins/double-space.plugin.ts
+++ b/projects/core/src/lib/plugins/double-space.plugin.ts
@@ -1,0 +1,74 @@
+import type {MaskitoPlugin} from '@maskito/core';
+
+import type {TypedInputEvent} from '../types';
+import {EventListener} from '../utils';
+
+const SPACE = ' ';
+
+/**
+ * 1. Android user (with G-board keyboard or similar) presses 1st space
+ * ```
+ * {type: "beforeinput", data: " ", inputType: "insertText"}
+ * ```
+ * 2. User presses 2nd space
+ * ```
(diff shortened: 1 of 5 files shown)


projects/core/src/lib/plugins/double-space.plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+ * ```
+ * ---------
+ * 1. MacOS user presses 1st space
+ * ```
+ * {type: "beforeinput", data: " ", inputType: "insertText"}
+ * ```
+ * 2. User presses 2nd space
+ * ```
+ * // MacOS automatically run `element.setSelectionRange(indexBeforeSpace, indexAfterSpace)` and then
+ * {type: "beforeinput", data: " ", inputType: "insertText"}
```
waterplea, line 27: ". "

waterplea (approved):

PR #2053, opened 2025-04-11
fix(core): add possibility to overwrites `selection` property in processors

### LINKED ISSUE #2051: Need help with preprocessors
I need a numeric input (full numbers and decimal numbers) that overwrites the 0 if the cursor is in the first position.
So for 0,234, if the cursor is in the first position and i press 1, the number should be 1,234 and the cursor should be after 1.
Same for integer inputs with value 0. If i type 1 i want 1 as a value with the cursor after 1.

I looked into preprocessors but as far as i can tell i can only change the data property in there.
I would need to update the selection to make this work.
So i would select [0,1] in both cases so the 0 gets replaced.

Am i missing something here or do i need to find a different approach?

nsbarsukov:
Fixes #2051

### DIFF
--- a/projects/core/src/lib/mask.ts
+++ b/projects/core/src/lib/mask.ts
@@ -306,7 +306,7 @@ export class Maskito extends MaskHistory {
             return event.preventDefault();
         }
 
-        const [from, to] = elementState.selection;
+        const [from, to] = initialElementState.selection;
         const newPossibleValue =
             initialElementState.value.slice(0, from) +
             data +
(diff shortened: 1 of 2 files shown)


projects/core/src/lib/mask.ts
```diff
@@ -306,7 +306,7 @@ export class Maskito extends MaskHistory {
             return event.preventDefault();
         }
 
-        const [from, to] = elementState.selection;
+        const [from, to] = initialElementState.selection;
         const newPossibleValue =
             initialElementState.value.slice(0, from) +
```
nsbarsukov, line 311: Fix only logic inside `handleInsert` because `handleDelete` already has the valid one

https://github.com/taiga-family/maskito/blob/218789d643de5da1f2d29accc50ccd8406a4f567/projects/core/src/lib/mask.ts#L268-L270

waterplea (approved):

PR #2087, opened 2025-04-25
fix(kit): `Number` should support non-erasable minus (as `prefix`) for `max <= 0`

nsbarsukov:
Relates:
* https://github.com/taiga-family/taiga-ui/issues/10910

### DIFF
--- a/projects/demo-integrations/src/tests/kit/number/number-prefix-postfix.cy.ts
+++ b/projects/demo-integrations/src/tests/kit/number/number-prefix-postfix.cy.ts
@@ -1,14 +1,18 @@
-import {DemoPath} from '@demo/constants';
+import {
+    CHAR_EM_DASH,
+    CHAR_EN_DASH,
+    CHAR_HYPHEN,
+    CHAR_MINUS,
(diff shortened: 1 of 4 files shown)


projects/demo-integrations/src/tests/kit/number/number-prefix-postfix.cy.ts
```diff
@@ -1,14 +1,18 @@
-import {DemoPath} from '@demo/constants';
+import {
+    CHAR_EM_DASH,
+    CHAR_EN_DASH,
+    CHAR_HYPHEN,
+    CHAR_MINUS,
+} from 'projects/kit/src/lib/constants';
```
splincode, line 6: `projects/` I don't remember, is it okey path for tests?

nsbarsukov: We have some problems with playwright (Taiga UI project) but it is okay for Cypress (it works locally and in CI)

splincode (approved):

vladimirpotekhin (approved):

PR #2090, opened 2025-04-25
fix(core): incorrect handle of paste event for `<input />` with `maxlength` attribute

### LINKED ISSUE #1993: 🐞 - Incorrect handle of paste event for `<input />` with `maxlength` attribute

@maskito/core

### Playground Link

https://stackblitz.com/edit/max-length-paste-bug

### Description

```html
<input maxlength="3" placeholder="Paste number with >3 digits" />
```

Paste
```
123456789
```

#### Without Maskito

Textfield contains `123`.

#### With Maskito

**Expected behavior**: textfield contains `123`.
**Actual behavior**: textfield is empty.

### Maskito version

3.3.0

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
Fixes #1993

### DIFF
--- a/projects/core/src/lib/mask.ts
+++ b/projects/core/src/lib/mask.ts
@@ -318,14 +318,12 @@ export class Maskito extends MaskHistory {
                 initialElementState.value.slice(to),
             selection: [from + data.length, from + data.length],
         };
-        const newElementState = this.postprocessor(maskModel, initialElementState);
-
-        if (newElementState.value.length > maxLength) {
-            return event.preventDefault();
-        }
+        const newElementState = this.clampState(
+            this.postprocessor(maskModel, initialElementState),
+        );
 
         if (
-            !areElementStatesEqual(newPossibleState, newElementState) ||
+            !areElementStatesEqual(this.clampState(newPossibleState), newElementState) ||
             element.isContentEditable
         ) {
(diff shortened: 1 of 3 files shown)


projects/core/src/lib/mask.ts
```diff
@@ -318,14 +318,12 @@ export class Maskito extends MaskHistory {
                 initialElementState.value.slice(to),
             selection: [from + data.length, from + data.length],
```
splincode, line 319: why not 

```ts
const newPossibleState: ElementState = this.clampState({
            value:
                initialState.value.slice(0, initialFrom) +
                initialState.value.slice(initialTo),
            selection: [initialFrom, initialFrom],
        });
```

splincode (approved):

waterplea (approved):

projects/core/src/lib/mask.ts
nsbarsukov: Variable `newPossibleState` is used twice in this method.
The last usage:
```js
newPossibleState.value.length > maxLength
```

We should use **un**clampled version for this check.

PR #2153, opened 2025-06-02
feat(react): add support for React-specific `onChange` event handler

nsbarsukov:
This PR solves 2 problems

## Problem 1
```jsx
const options: MaskitoOptions = {
    mask: /^[a-z]+$/i,
    postprocessors: [({value, selection}) => ({selection, value: value.toUpperCase()})],
};

function App() {
    const inputRef = useMaskito({options});
    const [value, neverCalledFn] = useState('');

    return (
        <input
            ref={inputRef}
            value={value}
            onInput={(e) => console.log(e.target.value)}
        />
    );
}
```

**Expected behavior:** textfield value is always empty
**Actual behavior:** textfield value is updated on every keystroke

Learn more: https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable

<br />

## Problem 2
```jsx
const options: MaskitoOptions = {
    mask: /^[a-z]+$/i,
    postprocessors: [({value, selection}) => ({selection, value: value.toUpperCase()})],
};

function App() {
    const inputRef = useMaskito({options});

    return (
        <input
            ref={inputRef}
            onChange={(e) => console.log(e.target.value)}
        />
    );
}
```

**Expected behavior:** `onChange` event is never be emitted, no console logs. 
**Actual behavior:** `onChange` event is emitted on every keystroke

Learn more: https://maskito.dev/frameworks/react#controlled-input

___

## Possible solutions

1. Inspired by https://github.com/facebook/react/issues/11488#issuecomment-347775628

<details>
<summary>Working solution via <code>_valueTracker</code> property </summary>

```ts
import type {MaskitoElement} from '@maskito/core';

/**
 * React adds `_valueTracker` property to every textfield elements for its internal logic with controlled inputs.
 * Also, React monkey-patches `value`-setter of the native textfield elements to update state inside its `_valueTracker`.
 * @see https://github.com/facebook/react/blob/ee76351917106c6146745432a52e9a54a41ee181/packages/react-dom-bindings/src/client/inputValueTracking.js#L12-L19
 */
type ReactControlledElement = MaskitoElement & {
    _valueTracker: {
        getValue(): string;
        setValue(value: string): void;
        stopTracking(): void;
    };
};

const isReactControlledElement = (
    element: MaskitoElement,
): element is ReactControlledElement => '_valueTracker' in element;

export function adaptReactControlledElement(element: MaskitoElement): MaskitoElement {
    if (!isReactControlledElement(element)) {
        return element;
    }

    const adapter = {
        set value(value: string) {
            const lastValue = element._valueTracker.getValue();

            element.value = value; // It will also update `_valueTracker` state
            /**
             * React depends on `_valueTracker` to know if the value was changed to decide:
             * - should it revert state for controlled input (if its state handler does not update value)
             * - should it dispatch its synthetic (not native!) `change` event
             * ___
             * When Maskito patches textfield with a valid value (using setter of `value` property),
             * it also updates `_valueTracker` state and React mistakenly decides that nothing has happened.
             * React should update `_valueTracker` state by itself.
             * ___
             * @see https://github.com/facebook/react/blob/ee76351917106c6146745432a52e9a54a41ee181/packages/react-dom-bindings/src/client/inputValueTracking.js#L173-L177
             */
            element._valueTracker.setValue(lastValue);
        },
    };

    return new Proxy(element, {
        get(target, prop: keyof HTMLElement) {
            const nativeProperty = target[prop];

            return typeof nativeProperty === 'function'
                ? nativeProperty.bind(target)
                : nativeProperty;
        },
        set(target, prop: keyof HTMLElement, val, receiver) {
            return Reflect.set(prop in adapter ? adapter : target, prop, val, receiver);
        },
    });
}

```
</details>

2. Cypress solution (see this [comment](https://github.com/cypress-io/cypress/issues/647#issuecomment-335829482) and [this PR](https://github.com/cypress-io/cypress/pull/732/files#diff-53b7d619fa4613db14158a57ca69e8170a6f5e3d2271c61e762c801bff15552dR33-R40)):

> My idea is this - when we go to change the .value= of an input, we first check to see if that property has been overwritten using Object.getOwnPropertyDescriptor. If we reset it to
HTMLInputElement.prototype.value or even just call the setter function directly with the input as context, then react will not be notified of the value change. That'll end up setting the value and bypass the overridden method - which mimics exactly what happens when a browser silently changes the value property.

This solution was introduced in Cypress@1.0.2 and [still exists in Cypress@14](https://github.com/cypress-io/cypress/blob/a0a4eb1b2c1baac9f4d6ad4b02a3db44b4abd70b/packages/driver/src/dom/elements/nativeProps.ts#L42-L44).

This PR chooses Cypress-way to solve this same issue.

### DIFF
--- a/projects/demo/src/pages/frameworks/react/react-doc.template.html
+++ b/projects/demo/src/pages/frameworks/react/react-doc.template.html
@@ -82,7 +82,7 @@
             <strong>Maskito</strong>
             too but usage of
             <code>onInput</code>
-            handler is more robust solution!
+            handler is a more robust solution!
         </tui-notification>
 
         <tui-doc-code [code]="controlledInputDemo" />
--- a/projects/demo/src/pages/frameworks/react/react-doc.template.html
+++ b/projects/demo/src/pages/frameworks/react/react-doc.template.html
@@ -55,11 +55,13 @@
         <p>
             <strong>Maskito</strong>
             core is developed as framework-agnostic library. It does not depend on any JS-framework's peculiarities. It
-            uses only native browser API. That is why you should use native
+            uses only native browser API. That is why we
+            <u>strongly recommends</u>
+            use native
             <code>onInput</code>
             instead of React-specific
             <code>onChange</code>
-            event. Do not worry, both events works similarly! Read more about it in the
+            event handler. Do not worry, both events works similarly! Read more about it in the
             <a
                 href="https://react.dev/reference/react-dom/components/input#props"
                 rel="noreferrer"
@@ -69,6 +71,19 @@
                 official React documentation.
             </a>
         </p>
+
+        <tui-notification
+            size="m"
+            class="tui-space_bottom-4"
+        >
+            React-specific
+            <code>onChange</code>
+            is supported by
+            <strong>Maskito</strong>
+            too but usage of
+            <code>onInput</code>
+            handler is more robust solution!
+        </tui-notification>
 
         <tui-doc-code [code]="controlledInputDemo" />
     </section>
--- a/projects/demo-integrations/src/tests/component-testing/react/change-event/change-event.cy.tsx
+++ b/projects/demo-integrations/src/tests/component-testing/react/change-event/change-event.cy.tsx
@@ -1,0 +1,148 @@
+import type {MaskitoOptions} from '@maskito/core';
+import {maskitoNumberOptionsGenerator} from '@maskito/kit';
+import {useMaskito} from '@maskito/react';
+import {mount} from 'cypress/react';
+import type {ChangeEvent, JSX} from 'react';
+import {useCallback, useState} from 'react';
+
+describe('React synthetic "onChange" event', () => {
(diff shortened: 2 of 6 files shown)


vladimirpotekhin (approved):

waterplea (approved):

projects/demo/src/pages/frameworks/react/react-doc.template.html
```diff
@@ -70,6 +72,19 @@ <h2>Controlled masked input</h2>
...
+            size="m"
+            class="tui-space_bottom-4"
+        >
+            React-specific
+            <code>onChange</code>
+            is supported by
+            <strong>Maskito</strong>
+            too but usage of
+            <code>onInput</code>
+            handler is more robust solution!
```
nsbarsukov:
```suggestion
            handler is a more robust solution!
```

vladimirpotekhin (approved):

PR #2154, opened 2025-06-03
chore(demo-integrations): enable React Cypress Component Testing

nsbarsukov:
```tsx
import {mount} from 'cypress/react';

const options: MaskitoOptions = {mask: /^\d+$/};

export const App: ComponentType = () => {
    const maskRef = useMaskito({options});

    return <input ref={maskRef} />;
};

describe('React component testing', () => {
    beforeEach(() => {
        mount(<App />);
    });

    it('works now', () => {
        cy.get('input').type('12abc3').should('have.value', '123');
    });

    // [...]
});

```


.github/renovate.json
```diff
@@ -15,7 +15,7 @@
     },
     "packageRules": [
         {
-            "matchPackageNames": ["/^@nx.*/", "/^nx$/"],
+            "matchPackageNames": ["/^@nx.*/", "/^nx$/", "cypress"],
```
nsbarsukov, line 18: `@nx/cypress` dep has its own restrictions about cypress version (Cypress should not be updated automatically and should be sync with Nx dep)

projects/demo/package.json
```diff
@@ -30,8 +30,8 @@
         "@taiga-ui/styles": "4.39.2",
         "@taiga-ui/testing": "4.39.2",
         "ngx-highlightjs": "10.0.0",
-        "react": "19.1.0",
-        "react-dom": "19.1.0",
+        "react": "18.3.1",
+        "react-dom": "18.3.1",
```
nsbarsukov, line 34: It will be 19th again later in next PR (I need bump nx, then bump cypress)

https://docs.cypress.io/app/references/changelog#14-0-0

nx.json
```diff
@@ -90,6 +90,10 @@
         "component-test": {
             "cache": true,
             "inputs": ["default", "^production"]
+        },
+        "ct-react": {
```
waterplea, line 94: Should we rename original one to ct-angular?

nsbarsukov: I've already tried to to it.
But there is a bug in `@nx/angular`. When i change default naming to `ct-angular`, it breaks react one:

```
➜ nx ct-react demo-integrations  

Your configFile is invalid: /Users/n.barsukov/WebstormProjects/maskito/projects/demo-integrations/cypress-react.config.ts

It threw an error when required, check the stack trace below:

Error: Unable to find component testing target configuration in project 'demo-integrations'.
Has targets? true
Has target name? component-test
Has ct project name? demo-integrations
    at getBuildableTarget (~/maskito/node_modules/@nx/angular/plugins/component-testing.js:95:15)
                                                         ⬆️ 
```

I will try to it later with upgrading of Nx.

Issue #2183, opened 2025-07-02
🚀 - `Time` supports `prefix` / `postfix` parameters


nsbarsukov:
### Which package(s) are relevant/related to the feature request?

@maskito/kit

### Description

Relates:
* https://github.com/taiga-family/taiga-ui/issues/11199

PR #2184, opened 2025-07-02
fix(core): invalid behavior of dynamic mask expression with trailing fixed characters

nsbarsukov:
Relates:
* https://github.com/taiga-family/maskito/issues/2183

### DIFF
--- a/projects/core/src/lib/classes/mask-model/tests/mask-model-fixed-characters.spec.ts
+++ b/projects/core/src/lib/classes/mask-model/tests/mask-model-fixed-characters.spec.ts
@@ -52,7 +52,7 @@ describe('MaskModel | Fixed characters', () => {
             );
 
             try {
-                maskModel.addCharacters(selection, addedCharacters);
+                maskModel.addCharacters(addedCharacters);
             } finally {
                 expect(maskModel.value).toBe(expectedNewValue);
                 expect(maskModel.selection).toEqual([
@@ -124,7 +124,7 @@ describe('MaskModel | Fixed characters', () => {
                 const selection = [2, 2] as const;
                 const maskModel = new MaskModel({value, selection}, options);
 
-                expect(() => maskModel.addCharacters(selection, 'q')).toThrow();
+                expect(() => maskModel.addCharacters('q')).toThrow();
                 expect(maskModel.value).toBe(value);
                 expect(maskModel.selection).toEqual(selection);
             });
@@ -134,7 +134,7 @@ describe('MaskModel | Fixed characters', () => {
                 const selection = [1, 1] as const;
                 const maskModel = new MaskModel({value, selection}, options);
 
-                expect(() => maskModel.addCharacters(selection, '$')).toThrow();
+                expect(() => maskModel.addCharacters('$')).toThrow();
                 expect(maskModel.value).toBe(value);
                 expect(maskModel.selection).toEqual(selection);
             });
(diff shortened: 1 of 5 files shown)


projects/kit/src/lib/processors/tests/valid-date-preprocessor.spec.ts
```diff
@@ -40,4 +40,25 @@ describe('createValidDatePreprocessor', () => {
...
+        const initialState = {
+            value: '06.02.2023',
+            selection: ['06.02.2023'.length, '06.02.2023'.length] as const,
+        };
+        const {elementState, data} = processor(
+            {elementState: initialState, data: rangeSeparator},
+            'insert',
+        );
+
+        expect(elementState).toEqual(initialState);
```
nsbarsukov, line 61: Without new code changes this test fails with the following error:
```
- Expected  - 1
+ Received  + 1

  Object {
    "selection": Array [
      10,
      10,
    ],
-   "value": "06.02.2023",
+   "value": "06.02.2023000",
  }
```

projects/core/src/lib/classes/mask-model/tests/mask-model-fixed-characters.spec.ts
```diff
@@ -144,13 +144,107 @@ describe('MaskModel | Fixed characters', () => {
...
+                {
+                    selection: [1, 1],
+                    value: '12:3left',
+                },
+                timeWithPostfixMask,
+            );
+
+            maskModel.addCharacters('4');
+
+            expect(maskModel.value).toBe('14:3left');
```
nsbarsukov, line 213: Without new code changes this test fails with the following error:

```
Expected: "14:3left"
Received: "14:3"
```

projects/core/src/lib/classes/mask-model/tests/mask-model-fixed-characters.spec.ts
```diff
@@ -144,13 +144,107 @@ describe('MaskModel | Fixed characters', () => {
...
+                    selection: [3, 4],
+                    value: '12:3left',
+                },
+                timeWithPostfixMask,
+            );
+
+            maskModel.deleteCharacters();
+
+            expect(maskModel.value).toBe('12left');
+            expect(maskModel.selection).toEqual([2, 2]);
```
nsbarsukov, line 229: Without new code changes this test fails with the following error:

```
- Expected  - 2
+ Received  + 2

  Array [
-   2,
-   2,
+   6,
+   6,
  ]
```

projects/core/src/lib/mask.ts
```diff
@@ -268,15 +268,16 @@ export class Maskito extends MaskHistory {
...
 
-        maskModel.deleteCharacters([from, to]);
+        maskModel.deleteCharacters();
 
         const newElementState = this.postprocessor(maskModel, initialState);
 
         if (
             areElementValuesEqual(initialState, elementState, maskModel, newElementState)
         ) {
+            const [from, to] = elementState.selection;
```
waterplea, line 279: The reason for moving is to get them after characters are already deleted?

waterplea (approved):

projects/core/src/lib/mask.ts
nsbarsukov: This line was moved because it is only used inside this `if`-block.
Honestly, no special reasons for it. Just move all connected variables close to each others.

`elementState.selection` is not mutated array (even before/after `maskModel.deleteCharacters` execution)

Issue #2255, opened 2025-08-14
🚀 - Support signal-based reactivity for @maskito/angular directive options without manual ngOnChanges() calls


OmerGronich:
### Which package(s) are relevant/related to the feature request?

@maskito/angular

### Description

We’re integrating `@maskito/angular` into a modern Angular setup that uses [signals](https://angular.dev/guide/signals) instead of classic `@Input()` bindings.

---

## The Problem

`MaskitoDirective` is currently built around the **classic Angular pattern**:

- Options are passed via `@Input() options: MaskitoOptions`
- Changes are handled in `ngOnChanges()`

When an `@Input()` changes via **template binding**, Angular calls `ngOnChanges()` and Maskito correctly re-creates its internal mask.

However, when **assigning `options` programmatically** from an `effect` in a host directive, Angular **does not** trigger `ngOnChanges()` (since it only fires for template-driven input changes).

This means the mask isn’t updated unless we do this:

```ts
effect(() => {
  this.maskito.options = { ... };
  this.maskito.ngOnChanges(); // manual lifecycle nudge
});
```

While this works, it’s a **compatibility shim**. Manually invoking lifecycle hooks is brittle, goes against Angular best practices, and leaks internal implementation details to the consumer.

---

## What Taiga UI Does

Taiga UI handles a similar integration via a [`tuiDirectiveBinding()`](https://github.com/taiga-family/taiga-ui/blob/main/projects/kit/utils/maskito.binding.ts) helper, which:

- Updates the directive’s property
- Explicitly calls `ngOnChanges()` internally

This is a **pragmatic shim**: it centralizes the manual lifecycle trigger so consumers don’t have to scatter `ngOnChanges()` calls across codebases. It still **emulates** the old lifecycle rather than providing signal-native reactivity, useful in practice, but ultimately something I would consider a hacky stopgap.

---

### A) Linked Signal + Effect  
**Demo:** https://stackblitz.com/edit/stackblitz-starters-6hmjpabr?file=src%2Ftime-input.directive.ts

- Expose `options` as a writable signal (`writableOptions`).
- Internally, use an `effect()` to destroy and re-create the `Maskito` instance whenever `options` or the element predicate changes.
- No reliance on `ngOnChanges`—directive becomes **reactive by construction**.

**Pros:** idiomatic modern Angular; no manual lifecycle calls.  
**Cons:** breaking change to directive API; requires migration for existing consumers.

---

### B) DI Override Pattern to Layer Custom Options  
[...]

PR #2257, opened 2025-08-15
fix(angular): use `@Input` setters instead of `ngOnChanges` to handle programmatic changes

OmerGronich:
Replaces `ngOnChanges`  in MaskitoDirective with `@Input` getters/setters for `options` and `elementPredicate`.

Why:
- Programmatic assignments do not trigger `ngOnChanges`, causing the mask not to update.
- Setters allow updates to be applied regardless of whether the change originates from template bindings or code.

Changes:
- Added `@Input` getter/setter pairs for `options` and `elementPredicate`.
- Preserved `ngOnChanges()` as a no-op for backwards compatibility.
- Renamed private fields and reordered members to satisfy ESLint rules.

Impact:
- Fixes mask not reinitializing on programmatic input changes.
- No breaking API changes for existing consumers.

**Before:**

```typescript
effect(() => {
  this.maskito.options = { ... };
  this.maskito.ngOnChanges(); // had to call manually to apply changes
});
```

**After:**

```typescript
effect(() => {
  this.maskito.options = { ... }; // setter applies changes automatically
});
```

Closes: https://github.com/taiga-family/maskito/issues/2255

### DIFF
--- a/projects/angular/src/lib/maskito.directive.ts
+++ b/projects/angular/src/lib/maskito.directive.ts
@@ -1,5 +1,14 @@
 import type {OnChanges, OnDestroy} from '@angular/core';
-import {Directive, ElementRef, inject, Input, NgZone} from '@angular/core';
+import {
+    Directive,
+    effect,
+    ElementRef,
+    inject,
+    Input,
+    NgZone,
+    signal,
+    untracked,
+} from '@angular/core';
 import {DefaultValueAccessor} from '@angular/forms';
 import type {MaskitoElementPredicate, MaskitoOptions} from '@maskito/core';
 import {
@@ -14,9 +23,35 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
     private readonly ngZone = inject(NgZone);
     private maskedElement: Maskito | null = null;
 
-    private optionsInternal: MaskitoOptions | null = null;
-    private elementPredicateInternal: MaskitoElementPredicate =
-        MASKITO_DEFAULT_ELEMENT_PREDICATE;
+    // TODO(v4): delete excessive setters/getters
+    private readonly optionsInternal = signal<MaskitoOptions | null>(null);
+    private readonly elementPredicateInternal = signal<MaskitoElementPredicate>(
+        MASKITO_DEFAULT_ELEMENT_PREDICATE,
+    );
+
+    protected readonly initEffect = effect(async () => {
+        const elementPredicate = this.elementPredicateInternal();
+        const options = this.optionsInternal();
+        const {maskedElement, elementRef, ngZone} = this;
+
+        maskedElement?.destroy();
+
+        if (!options) {
+            return;
+        }
+
+        const predicateResult = await elementPredicate(elementRef);
+
+        if (this.elementPredicate !== elementPredicate || this.options !== options) {
+            // Ignore the result of the predicate if the
+            // maskito element (or its options) has changed before the predicate was resolved.
+            return;
+        }
+
+        ngZone.runOutsideAngular(() => {
+            this.maskedElement = new Maskito(predicateResult, options);
+        });
+    });
 
     constructor() {
         const accessor = inject(DefaultValueAccessor, {self: true, optional: true});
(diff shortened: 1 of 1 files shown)


projects/angular/src/lib/maskito.directive.ts
```diff
@@ -28,37 +26,77 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
...
+        this.elementPredicateInternal = value;
+        void this.setup();
+    }
+
+    // eslint-disable-next-line @angular-eslint/no-empty-lifecycle-method
     public async ngOnChanges(): Promise<void> {
-        const {elementPredicate, options, maskedElement, elementRef, ngZone} = this;
+        /**
+         * no-op for backwards compatibility.
+         */
```
nsbarsukov:
```suggestion
        // TODO(v4): delete
```

projects/angular/src/lib/maskito.directive.ts
```diff
@@ -28,37 +26,77 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
...
+    public ngOnDestroy(): void {
+        this.maskedElement?.destroy();
+    }
+
+    private async setup(): Promise<void> {
+        const options = this.optionsInternal;
+        const elementPredicate = this.elementPredicateInternal;
+
+        this.maskedElement?.destroy();
+        this.maskedElement = null;
```
nsbarsukov: Do we really need this line ?

projects/angular/src/lib/maskito.directive.ts
```diff
@@ -28,37 +26,77 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
...
         }
     }
 
+    @Input('maskito')
+    public get options(): MaskitoOptions | null {
+        return this.optionsInternal;
+    }
+
+    @Input('maskitoElement')
+    public get elementPredicate(): MaskitoElementPredicate {
```
nsbarsukov: Let's put these decorators on setters instead of getters

```ts
    @Input('maskito')
    public set options(value: MaskitoOptions | null) {
    // [...]

    @Input('maskitoElement')
    public set elementPredicate(value: MaskitoElementPredicate) {
    
    // [...]
```

projects/angular/src/lib/maskito.directive.ts
```diff
@@ -28,37 +26,77 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
...
+        return this.elementPredicateInternal;
+    }
+
+    public set options(value: MaskitoOptions | null) {
+        if (this.optionsInternal === value) {
+            return;
```
nsbarsukov: I am not sure that we really need this check with early exit for new setters.

`MaskitoOptions` & `MaskitoElementPredicate` are both not-primitives. So, checking equality will return `false` for the same objects but with different references. 

As for `null` – input property will not be triggered if we `null` will be replaced by `null` (even without this check). 
The same happens with signal based solution

```ts
effect(() => {
  this.maskito.options = this.mask();
});
```

projects/angular/src/lib/maskito.directive.ts
```diff
@@ -28,37 +26,77 @@ export class MaskitoDirective implements OnDestroy, OnChanges {
...
+        if (this.elementPredicateInternal === value) {
+            return;
+        }
+
+        this.elementPredicateInternal = value;
+        void this.setup();
+    }
+
+    // eslint-disable-next-line @angular-eslint/no-empty-lifecycle-method
     public async ngOnChanges(): Promise<void> {
```
nsbarsukov, line 91: Thread to discuss you concern https://github.com/taiga-family/maskito/issues/2255#issuecomment-3191676552

> My only concern is that using setters could make the setup function run more often than needed, because it will be called whenever either options or elementPredicate is set, so if both are set, the setup function runs twice. This differs from the current ngOnChanges approach or an effect, which can handle both changes in a single call.

@taiga-family/core-team what do you think about it ?
What's worse:
- twice initialization of mask if predicate is used (current PR approach)
- unwanted microtask (angular signals + `effect` approach)
?

nsbarsukov: However, it should be mentioned that current implementation (using `ngOnChanges`) always produces extra microtask (even for sync predicate).

https://github.com/taiga-family/maskito/blob/1472840b5575cc8b9b16edbdd9be39ecb7bf1e32/projects/angular/src/lib/maskito.directive.ts#L39-L48

And we don't get any bug reports about it.
Maybe `effect` approach is okay ?

nsbarsukov: I suggest the following approach

```ts
// TODO(v4): use {@link model inputs https://angular.dev/guide/components/inputs#model-inputs} and delete excessive setters/getters
private readonly optionsInternal = signal<MaskitoOptions | null>(null);

@Input('maskito')
public set options(x: MaskitoOptions | null) {
    this.optionsInternal.set(x);
}

protected readonly initEffect = effect(() => {
    // this.optionsInternal()
    // this.elementPredicateInternal()
});
```

nsbarsukov: @OmerGronich Let's refactor PR according to my last comment.

In upcoming major release it will be replaced by model signals
* https://github.com/taiga-family/maskito/issues/2261

OmerGronich: > Let's refactor PR according to my last comment.

On it

> In upcoming major release it will be replaced by model signals

Model inputs don’t feel right here. They implicitly add an output to the API, but outputs are usually meant for notifying about user-driven interactions. In this case we just need to programmatically update a value from outside. I think `linkedSignal` is a better fit: 
https://stackblitz.com/edit/stackblitz-starters-6hmjpabr?file=src%2Ftime-input.directive.ts

I left out this part of the comment because of this: 

> use {@link model inputs https://angular.dev/guide/components/inputs#model-inputs} and

Let me know if you’d like me to include it anyway.

projects/angular/src/lib/maskito.directive.ts
OmerGronich: removed in refactor

projects/angular/src/lib/maskito.directive.ts
OmerGronich: removed in refactor

projects/angular/src/lib/maskito.directive.ts
nsbarsukov: > I think linkedSignal is a better fit

Maybe. However, `linkedSignal` was introduced only with Angular 19+. We don't have plans to bump Angular in our Taiga Family products to more than `18.x.y` in near future ([why?](https://medium.com/angularwave/introducing-taiga-ui-v4-df6c7c62330f#259e)).

> They implicitly add an output to the API

Yep, it is the main disadvantage of this approach.
However, `linkedSignal` approach also extends class `MaskitoDirective` (public property for input prop + public property for `linkedSignal` instead of single class property `public options: ModelSignal<MaskitoOptions>`).

nsbarsukov: > // TODO(v4): delete excessive setters/getters

This comment is enough 👍

Issue #2052, opened 2025-04-11
🐞 - `maskitoParseTime` / `maskitoStringifyTime` should support `AM` / `PM` formats


nsbarsukov:
### Which package(s) are the source of the bug?

@maskito/kit

### Playground Link

https://maskito.dev/kit/time

### Description

Relates:
* https://github.com/taiga-family/maskito/pull/1302
* https://github.com/taiga-family/maskito/pull/1708

> **Don't forget to add unit tests!**

### Maskito version

3.6.0

ggomarighetti:
Should adapt the operation of the utilities, trying to modify as little as possible the original code, adding helper variables at startup and modifying the return, and add the corresponding testing scenarios.

rashiddok:
I'll take it

PR #2260, opened 2025-08-18
fix(kit): `maskitoStringifyTime` and `maskitoParseTime` should support `AM` / `PM` formats

rashiddok:
Resume
Adapt the operation of the utilities, trying to modify as little as possible the original code, adding helper variables at startup and modifying the return, and add the corresponding testing scenarios.

Evidence
Not required

Reference
https://github.com/taiga-family/maskito/issues/2052


projects/kit/src/lib/masks/time/utils/stringify-time.ts
```diff
@@ -18,6 +18,7 @@ export function maskitoStringifyTime(
         ...timeSegmentMaxValues,
     };
 
+    const hasMeridiem = mode?.includes('AA');
```
nsbarsukov: Is optional chaining really required ?

https://github.com/taiga-family/maskito/blob/2d8c40b6ab728b4e58c2e4d08ae56b8ec8f33f21/projects/kit/src/lib/masks/time/time-params.ts#L3-L4

nsbarsukov (approved):

projects/kit/src/lib/masks/time/utils/stringify-time.ts
nsbarsukov:
```suggestion
    const hasMeridiem = mode.includes('AA');
```

PR #2281, opened 2025-08-25
feat(kit): `Number` supports minus before prefix

### LINKED ISSUE #360: 🚀 - add a flag to the maskitoNumberOptionsGenerator function that will place a minus sign before the prefix

_No response_

### Description

It could be helpful for currencies:
image
(should be −$2.35)

nsbarsukov:
### Previous behavior
https://github.com/taiga-family/maskito/blob/88d34075f26a71afb31460d31978c1fcb3b0cfc3/projects/kit/src/lib/masks/number/number-params.ts#L21

### New behavior
https://github.com/taiga-family/maskito/blob/bb6ea4974f78fe401fd7753f57e3500e40c6f230/projects/kit/src/lib/masks/number/number-params.ts#L21

```ts
const numberMask = maskitoNumberOptionsGenerator({
    minusSign: '-',
    prefix: ['-', '$'],
})
```

Fixes #360


projects/demo-integrations/src/tests/kit/number/number-minus-before-prefix.cy.ts
```diff
@@ -0,0 +1,213 @@
+import {CHAR_EM_DASH, CHAR_EN_DASH, CHAR_JP_HYPHEN} from 'projects/kit/src/lib/constants';
+
+import {openNumberPage} from './utils';
+
+describe('Number | Minus before prefix', () => {
+    describe('[prefix]=["-", "$"] | [minusSign]="-"', () => {
```
nsbarsukov:
```suggestion
    describe('[prefix]="$" & [minusSign]="-" & negativePattern="minusFirst"', () => {
```

projects/demo/src/pages/kit/number/number-mask-doc.template.html
```diff
@@ -143,10 +143,27 @@
...
+            [description]="minusBeforePrefixDescription"
+        >
+            <ng-template #minusBeforePrefixDescription>
+                <p>
+                    Use
+                    <code>negativePattern</code>
+                    property to configure order of prefix and minus sign (by default, prefix is always placed before
+                    minus).
+                </p>
+            </ng-template>
```
nsbarsukov:
```suggestion
            <ng-template #minusBeforePrefixDescription>
                Use
                <code>negativePattern</code>
                property to configure order of prefix and minus sign (by default, prefix is always placed before minus).
            </ng-template>
```

PR #2337, opened 2025-09-29
[WIP] fix(kit): `Number` has broken support for postfix with leading point

### LINKED ISSUE #2333: 🐞 - postfix starting with dot produces error

@maskito/kit, @maskito/core

### Playground Link

https://stackblitz.com/edit/uj5pre9y

### Description

After updating to `3.11.0` our number mask with prefix throws an error when removing everything.

I attached an stackblitz to reproduce:
1. Open
2. Click in the input
3. Either remove the "1" by backspace or select everything and then remove

You will see a error thrown.

I think it is because of the starting dot, because starting `-` for example works.

### Maskito version

3.11.0

- [ ] Chrome
- [ ] Firefox
- [x] Safari
- [ ] Edge

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
Fixes #2333


projects/demo-integrations/src/tests/kit/number/number-prefix-postfix.cy.ts
```diff
@@ -372,4 +372,192 @@ describe('Number | Prefix & Postfix', () => {
                 .should('have.prop', 'selectionEnd', '100'.length);
         });
     });
+
+    describe('postfix consists starts with point and contains digits ([postfix]=".000 km" & [maximumFractionDigits]="0")`', () => {
```
nsbarsukov:
```suggestion
    describe('postfix starts with point and contains digits ([postfix]=".000 km" & [maximumFractionDigits]="0")`', () => {
```

PR #2363, opened 2025-10-10
feat(angular)!: `MaskitoDirective` uses model inputs

### LINKED ISSUE #2261: 🚀 - Use model inputs for Angular `MaskitoDirective`

@maskito/angular

### New upcoming public API 
```ts
export class MaskitoDirective {
  public options: ModelSignal<MaskitoOptions | null>;
  public elementPredicate: ModelSignal<MaskitoElementPredicate>;
  public ngOnDestroy: () => void;
}
```

### Possible solution
https://stackblitz.com/edit/maskito-model-based-directive?file=src%2Fmaskito.directive.ts

```ts
@Directive({ standalone: true, selector: '[maskito]' })
export class MaskitoDirective implements OnDestroy {
  private readonly elementRef: HTMLElement = inject(ElementRef).nativeElement;
  private readonly ngZone = inject(NgZone);
  private maskedElement: Maskito | null = null;

  public options = model<MaskitoOptions | null>(null, { alias: 'maskito' });
  public elementPredicate = model<MaskitoElementPredicate>(
    MASKITO_DEFAULT_ELEMENT_PREDICATE,
    { alias: 'maskitoElement' }
  );

  constructor() {
    const accessor = inject(DefaultValueAccessor, {
      self: true,
      optional: true,
    });

    if (accessor) {
      const original = accessor.writeValue.bind(accessor);
      const options = this.options();

      accessor.writeValue = (value: unknown) => {
        original(
          options ? maskitoTransform(String(value ?? ''), options) : value
        );
      };
    }
  }

  protected initEffect = effect(async () => {
    const { maskedElement, elementRef, ngZone } = this;
    const options = this.options();
    const elementPredicate = this.elementPredicate();

    maskedElement?.destroy();

    if (!options) {
      return;
    }

    // TODO(v5): use Resouce API instead of async effect (Angular 19+)
    const predicateResult = await elementPredicate(elementRef);

    if (
      this.elementPredicate() !== elementPredicate ||
      this.options() !== options
    ) {
      // maskito element (or its options) has changed before the predicate was resolved.
      return;
    }

    ngZone.runOutsideAngular(() => {
      this.maskedElement = new Maskito(predicateResult, options);
    });
  });

  public ngOnDestroy(): void {
  }
}
```

nsbarsukov:
Fixes #2261

### DIFF
--- a/projects/angular/src/lib/pattern.directive.ts
+++ b/projects/angular/src/lib/pattern.directive.ts
@@ -1,16 +1,16 @@
-import {Directive, inject, Input} from '@angular/core';
+import {Directive, inject} from '@angular/core';
 
 import {MaskitoDirective} from './maskito.directive';
 
 @Directive({
     standalone: true,
     selector: '[maskitoPattern]',
+    inputs: ['maskitoPattern'],
     hostDirectives: [MaskitoDirective],
 })
 export class MaskitoPattern {
     private readonly maskitoDirective = inject(MaskitoDirective, {self: true});
 
-    @Input()
     public set maskitoPattern(pattern: RegExp | string) {
         this.maskitoDirective.options.set({
             mask: typeof pattern === 'string' ? new RegExp(`^${pattern}$`) : pattern,
--- a/projects/demo/src/pages/frameworks/angular/examples/5-custom-unmask-handler/unmask.directive.ts
+++ b/projects/demo/src/pages/frameworks/angular/examples/5-custom-unmask-handler/unmask.directive.ts
@@ -1,5 +1,5 @@
 import type {AfterViewInit} from '@angular/core';
-import {Directive, inject, Input} from '@angular/core';
+import {Directive, inject} from '@angular/core';
 import {DefaultValueAccessor} from '@angular/forms';
 import {MaskitoDirective} from '@maskito/angular';
 import {maskitoTransform} from '@maskito/core';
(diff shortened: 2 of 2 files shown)


projects/angular/src/lib/pattern.directive.ts
```diff
@@ -10,11 +10,10 @@ import {MaskitoDirective} from './maskito.directive';
 export class MaskitoPattern {
     private readonly maskitoDirective = inject(MaskitoDirective, {self: true});
 
-    @Input('maskitoPattern')
-    public set regExpStr(pattern: RegExp | string) {
-        this.maskitoDirective.options = {
+    @Input()
```
waterplea: Move it to `inputs` in `@Directive` to drop deprecated `@Input()` decorator.

projects/demo/src/pages/frameworks/angular/examples/5-custom-unmask-handler/unmask.directive.ts
```diff
@@ -17,10 +17,11 @@ export class UnmaskDirective implements AfterViewInit {
     public unmaskHandler: (value: string) => any = identity;
 
     @Input()
```
waterplea: Same here

projects/angular/src/lib/pattern.directive.ts
nsbarsukov: https://github.com/taiga-family/maskito/pull/2363/commits/27c15956fb2a9c7d77af28daf879f3f3d8811ce6

projects/demo/src/pages/frameworks/angular/examples/5-custom-unmask-handler/unmask.directive.ts
nsbarsukov: https://github.com/taiga-family/maskito/pull/2363/commits/27c15956fb2a9c7d77af28daf879f3f3d8811ce6

PR #2394, opened 2025-10-22
chore(demo): update `Custom unmask handler` example for `Angular` documentation page

### LINKED ISSUE #2391: 🐞 - ``Number`` fails when value is set programmatically

Don't known / other

### Playground Link

https://stackblitz.com/edit/maskito-angular-decimal-and-thousand?file=src%2Fmain.ts

### Description

When using the following config for MaskitoNumberParams:

```
  private readonly maskitoParams: MaskitoNumberParams = {
    minimumFractionDigits: 2,
    maximumFractionDigits: 2,
    thousandSeparator: '.',
    decimalSeparator: ',',
  };
```

The initial value ``1000.42`` is shown incorrectly as ``100.042,00`` instead of ``1.000,42``. The behavior is the same when the button is clicked, the patched value of ``1234567.89`` is displayed in the input as ``123.456.789,00`` instead of ``1.234.567,89``.

### Maskito version

4.0.0

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [x] Edge

- [ ] macOS
- [x] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
Fixes #2391

### DIFF
--- a/projects/demo-integrations/src/tests/angular/unmask-handler.cy.ts
+++ b/projects/demo-integrations/src/tests/angular/unmask-handler.cy.ts
@@ -2,7 +2,7 @@ import {DemoPath} from '@demo/constants';
 
 describe('Angular | Custom unmask handler', () => {
     beforeEach(() => {
-        cy.visit(`/${DemoPath.Angular}/API`);
+        cy.visit(DemoPath.Angular);
 
         cy.get('#unmask [automation-id="tui-doc-example"]').as('example');
         cy.get('@example').find('input').as('input');
--- a/projects/demo/src/pages/frameworks/angular/examples/5-custom-unmask-handler/index.ts
+++ b/projects/demo/src/pages/frameworks/angular/examples/5-custom-unmask-handler/index.ts
@@ -38,6 +38,7 @@ export class UnmaskDocExample5 {
     protected readonly unmaskHandler = (x: string): number =>
         maskitoParseNumber(x, NUMBER_PARAMS);
 
+    // `maskitoStringifyNumber` implements the reverse transformation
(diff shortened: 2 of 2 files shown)


projects/demo-integrations/src/tests/angular/unmask-handler.cy.ts
```diff
@@ -0,0 +1,32 @@
+import {DemoPath} from '@demo/constants';
+
+describe('Angular | Custom unmask handler', () => {
+    beforeEach(() => {
+        cy.visit(`/${DemoPath.Angular}/API`);
```
nsbarsukov:
```suggestion
        cy.visit(DemoPath.Angular);
```

projects/demo/src/pages/frameworks/angular/examples/5-custom-unmask-handler/index.ts
```diff
@@ -18,15 +29,15 @@ export class UnmaskDocExample5 {
...
 
     /**
      * `maskitoParseNumber` is built-in utility to convert
      * entered number (as prettified formatted STRING) to number-type value
      */
-    protected readonly unmaskHandler = maskitoParseNumber;
+    protected readonly unmaskHandler = (x: string): number =>
+        maskitoParseNumber(x, NUMBER_PARAMS);
+
+    protected readonly stringify = (x: number): string =>
```
nsbarsukov, line 42:
```suggestion
    // `maskitoStringifyNumber` implements the reverse transformation
    protected readonly stringify = (x: number): string =>
```

PR #2420, opened 2025-11-11
fix(kit): `Time` with `[step]` ignores `timeSegmentMinValues` on arrow stepping

### LINKED ISSUE #2346: 🐞 - `Time` with `[step]` ignores `timeSegmentMinValues` on arrow stepping

@maskito/react, @maskito/kit

### Playground Link

https://stackblitz.com/edit/vitejs-vite-rxdi8mxi?file=src%2FApp.tsx

### Description

timeSegmentMinValues not working when change time by arrows
Image

Image

Image

### Maskito version

3.10.1

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [x] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

collin-b1:
Fixes #2346

Hello! This should fix the issue where arrow key stepping time segments would ignore `timeSegmentMinValues` and only wrap to zero. Updated the `time-step.cy.ts` test to accurately reflect hour stepping for `HH:MM AA` mode (hour wraps from 12 -> 1).

I had some troubles getting the npm test scripts to pass even in main, so let me know if anything is not working 😄

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,7 @@
                 "@tinkoff/eslint-config": "5.0.1",
                 "@tinkoff/eslint-config-react": "5.0.1",
                 "@types/jest": "29.5.14",
-                "@types/node": "24.10.0",
+                "@types/node": "24.10.1",
                 "http-server": "14.1.1",
                 "husky": "9.1.7",
                 "jest": "29.7.0",
@@ -3384,9 +3384,9 @@
             "license": "MIT"
         },
         "node_modules/@bufbuild/protobuf": {
-            "version": "2.10.0",
-            "resolved": "https://registry.npmjs.org/@bufbuild/protobuf/-/protobuf-2.10.0.tgz",
-            "integrity": "sha512-fdRs9PSrBF7QUntpZpq6BTw58fhgGJojgg39m9oFOJGZT+nip9b0so5cYY1oWl5pvemDLr0cPPsH46vwThEbpQ==",
+            "version": "2.10.1",
+            "resolved": "https://registry.npmjs.org/@bufbuild/protobuf/-/protobuf-2.10.1.tgz",
+            "integrity": "sha512-ckS3+vyJb5qGpEYv/s1OebUHDi/xSNtfgw1wqKZo7MR9F2z+qXr0q5XagafAG/9O0QPVIUfST0smluYSTpYFkg==",
             "dev": true,
             "license": "(Apache-2.0 AND BSD-3-Clause)"
         },
@@ -3430,9 +3430,9 @@
             }
         },
         "node_modules/@cacheable/memory/node_modules/keyv": {
-            "version": "5.5.3",
-            "resolved": "https://registry.npmjs.org/keyv/-/keyv-5.5.3.tgz",
-            "integrity": "sha512-h0Un1ieD+HUrzBH6dJXhod3ifSghk5Hw/2Y4/KHBziPlZecrFyE9YOTPU6eOs0V9pYl8gOs86fkr/KN8lUX39A==",
+            "version": "5.5.4",
+            "resolved": "https://registry.npmjs.org/keyv/-/keyv-5.5.4.tgz",
+            "integrity": "sha512-eohl3hKTiVyD1ilYdw9T0OiB4hnjef89e3dMYKz+mVKDzj+5IteTseASUsOB+EU9Tf6VNTCjDePcP6wkDGmLKQ==",
             "dev": true,
             "license": "MIT",
             "peer": true,
@@ -3451,9 +3451,9 @@
             }
         },
         "node_modules/@cacheable/utils/node_modules/keyv": {
-            "version": "5.5.3",
-            "resolved": "https://registry.npmjs.org/keyv/-/keyv-5.5.3.tgz",
-            "integrity": "sha512-h0Un1ieD+HUrzBH6dJXhod3ifSghk5Hw/2Y4/KHBziPlZecrFyE9YOTPU6eOs0V9pYl8gOs86fkr/KN8lUX39A==",
+            "version": "5.5.4",
+            "resolved": "https://registry.npmjs.org/keyv/-/keyv-5.5.4.tgz",
+            "integrity": "sha512-eohl3hKTiVyD1ilYdw9T0OiB4hnjef89e3dMYKz+mVKDzj+5IteTseASUsOB+EU9Tf6VNTCjDePcP6wkDGmLKQ==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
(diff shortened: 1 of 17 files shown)


projects/kit/src/lib/masks/date-time/date-time-mask.ts
```diff
@@ -125,6 +125,7 @@ export function maskitoDateTimeOptionsGenerator({
             createTimeSegmentsSteppingPlugin({
                 step: timeStep,
                 fullMode,
+                timeSegmentMinValues: DEFAULT_TIME_SEGMENT_MIN_VALUES,
```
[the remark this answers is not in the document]
nsbarsukov: @collin-b1
Probably, AI is right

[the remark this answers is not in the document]
nsbarsukov: Add Cypress tests for `DateTime` too to cover this case

projects/kit/src/lib/plugins/time/time-segments-stepping.ts
```diff
@@ -110,10 +115,8 @@ function updateSegmentValue({
...
 
-function mod(value: number, max: number): number {
-    if (value < 0) {
-        value += Math.floor(Math.abs(value) / max + 1) * max;
-    }
+function mod(value: number, min: number, max: number): number {
+    const range = max - min;
 
-    return value % max;
+    return ((((value - min) % range) + range) % range) + min;
```
nsbarsukov, line 121: Do we really need to take reminder twice if `value >= 0` & `range > 0` ?

```
(X % Y) + Y) % Y === X % Y

where 
X >= 0
Y > 0
```

nsbarsukov:
@collin-b1 write anything inside this issue (e.g. `I'll take it`)
* https://github.com/taiga-family/maskito/issues/2346

(without your first message inside it, I cannot assign it to you – GitHub restriction)

projects/kit/src/lib/plugins/time/time-segments-stepping.ts
collin-b1:
```ts
function mod(value, min, max) {
    const range = max - min;

    return ((value - min) % range) + min;
}
// mod(0, 1, 13) => 0
```

```ts
function mod(value, min, max) {
    const range = max - min;

    return ((((value - min) % range) + range) % range) + min;
}
// mod(0, 1, 13) => 12
```
Shortened version works when `value >= min` but doesn't stay in `[min, max)` otherwise.

projects/demo-integrations/src/tests/kit/date-time/date-time-time-step.cy.ts
```diff
@@ -147,4 +147,75 @@ describe('DateTime | timeStep', () => {
...
+                    .should('have.a.prop', 'selectionEnd', '22.12;'.length)
+                    .invoke('val')
+                    .should('match', /^22\.12;02:34\sPM$/);
+
+                cy.get('@input')
+                    .type('{downArrow}'.repeat(4))
+                    .should('have.a.prop', 'selectionStart', '22.12;'.length)
+                    .should('have.a.prop', 'selectionEnd', '22.12;'.length)
+                    .invoke('val')
+                    .should('match', /^22\.12;10:34\sPM$/);
```
collin-b1: Did regex matching because have.value would inconsistently not work due to non-breaking spaces vs regular spaces.

projects/kit/src/lib/masks/date-time/date-time-mask.ts
```diff
@@ -125,8 +125,8 @@ export function maskitoDateTimeOptionsGenerator({
             createTimeSegmentsSteppingPlugin({
                 step: timeStep,
                 fullMode,
-                timeSegmentMinValues: DEFAULT_TIME_SEGMENT_MIN_VALUES,
-                timeSegmentMaxValues: DEFAULT_TIME_SEGMENT_MAX_VALUES,
```
collin-b1, line 128: Both of these needed to be updated to support proper min/max stepping with meridiem.

projects/kit/src/lib/plugins/time/time-segments-stepping.ts
nsbarsukov: Thanks for clarification!
Let's leave it as is

projects/demo-integrations/src/tests/kit/date-time/date-time-time-step.cy.ts
nsbarsukov: You could just directly use non-breaking space here

In the same way as here:
https://github.com/taiga-family/maskito/blob/7a51702361237a41cd9bbdcdbb8e46d0bfa2e4bc/projects/demo-integrations/src/tests/kit/time/time-meridiem.cy.ts#L20

nsbarsukov: And replace all occurrences of
```ts
.invoke('val')
.should('match', /.../);
```

with 
```ts
.should('have.value', '...')
```

projects/demo-integrations/src/tests/kit/date-time/date-time-time-step.cy.ts
```diff
@@ -147,4 +147,75 @@ describe('DateTime | timeStep', () => {
...
+                    .type('1234p')
+                    .invoke('val')
+                    .should('match', /^22\.12;12:34\sPM$/);
+
+                cy.get('@input')
+                    .should('have.a.prop', 'selectionStart', '22.12;12:34 PM'.length)
+                    .should('have.a.prop', 'selectionEnd', '22.12;12:34 PM'.length);
+
+                cy.get('@input')
+                    .type('{home}')
```
nsbarsukov:
```suggestion
                    .type('{moveToStart}')
```

(it is just more self-descriptive way to do the same)
https://docs.cypress.io/api/commands/type#Arguments

projects/demo-integrations/src/tests/kit/date-time/date-time-time-step.cy.ts
```diff
@@ -147,4 +147,75 @@ describe('DateTime | timeStep', () => {
...
+            it('increments and decrements minutes in AM/PM mode correctly', () => {
+                cy.get('@input').type('1234a');
+                cy.get('@input')
+                    .invoke('val')
+                    .should('match', /^22\.12;12:34\sAM$/);
+
+                cy.get('@input')
+                    .should('have.a.prop', 'selectionStart', '22.12;12:34 AM'.length)
+                    .should('have.a.prop', 'selectionEnd', '22.12;12:34 AM'.length)
+                    .type('{leftArrow}{leftArrow}{leftArrow}{upArrow}{upArrow}');
```
nsbarsukov:
```suggestion
                    .type('{leftArrow}'.repeat(3))
                    .type('{upArrow}'.repeat(2));
```

nsbarsukov (approved):

PR #2507, opened 2026-01-21
fix(react): `useMaskito` should destroy Maskito instance if element is detached from the DOM

### LINKED ISSUE #2505: 🐞 - useMaskito not properly destroying if input gets destroyed

@maskito/react

### Playground Link

https://stackblitz.com/edit/vitejs-vite-oirkaspz?file=package.json,src%2FApp.tsx&terminal=dev

### Description

**Description**
When an input using `useMaskito` is conditionally rendered (e.g. toggled on/off), masking works on the first render, **fails on the second render**, works on the third, fails on the fourth, etc. The behavior alternates. See the stackblitz. At least Maskito version 3 and 5 are affected (I assume version 4 and previous versions are affected too).

**Expected behavior**
Masking should apply every time the input is rendered.

**Fix**
A working fix in our case is to `setElement` to `null` in the `useIsomorphicLayoutEffect` like:

https://github.com/taiga-family/maskito/blob/cd73d6a729068e4b301509b12def6e149a7b5d66/projects/react/src/lib/useMaskito.ts#L54-L57

```ts
useIsomorphicLayoutEffect(() => {
    if (!hostElement) {
      setElement(null); // fixes the issue
      return;
    }
```

### Maskito version

5.0.1

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [ ] Windows
- [x] Linux
- [ ] iOS
- [ ] Android

HagenMorano:
Fixes #2505

### DIFF
--- a/projects/react/src/lib/useMaskito.ts
+++ b/projects/react/src/lib/useMaskito.ts
@@ -70,6 +70,10 @@ export const useMaskito = ({
         } else {
             setElement(elementOrPromise);
         }
+
+        return () => {
+            setElement(null);
+        };
     }, [hostElement, elementPredicate, latestPredicateRef, options, latestOptionsRef]);
 
     useIsomorphicLayoutEffect(() => {
--- a/projects/react/src/lib/tests/useMaskito.spec.tsx
+++ b/projects/react/src/lib/tests/useMaskito.spec.tsx
@@ -1,6 +1,7 @@
 import type {MaskitoOptions} from '@maskito/core';
+import * as maskitoCore from '@maskito/core';
 import type {RenderResult} from '@testing-library/react';
-import {render} from '@testing-library/react';
+import {render, waitFor} from '@testing-library/react';
 import type {UserEvent} from '@testing-library/user-event';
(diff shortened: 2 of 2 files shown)


projects/react/src/lib/useMaskito.ts
```diff
@@ -53,6 +53,8 @@ export const useMaskito = ({
 
     useIsomorphicLayoutEffect(() => {
         if (!hostElement) {
+            setElement(null);
```
nextZed: It does the trick, although the more React way would be to return a cleanup function:

```ts
    useIsomorphicLayoutEffect(() => {
        if (!hostElement) {
            return;
        }

        const elementOrPromise = elementPredicate(hostElement);

        if (isThenable(elementOrPromise)) {
          // ...
        }

        return () => {
            setElement(null); // <- here
        };
    }, [hostElement, elementPredicate, latestPredicateRef, options, latestOptionsRef]);
```

HagenMorano: Agree and force pushed the change :+1:

PR #2528, opened 2026-02-05
chore: simplify serve

splincode:


projects/demo/project.json
```diff
@@ -90,19 +90,7 @@
...
-                        "@taiga-ui/cdk",
-                        "@taiga-ui/core",
-                        "@taiga-ui/icons",
-                        "@taiga-ui/kit",
-                        "@taiga-ui/layout",
-                        "@taiga-ui/legacy",
-                        "@taiga-ui/styles"
-                    ]
-                }
+                "prebundle": false
```
nsbarsukov, line 93: Why do we need to disable this feature for all the third-party project dependencies ?

https://angular.dev/tools/cli/build-system-migration#prebundling

nsbarsukov: > Prebundle removes the need to rebuild and bundle the project's dependencies each time a rebuild occurs or the development server is executed."

splincode: I didn’t notice a significant drop in build speed; on the upside, you don’t have to specify all the packages here.

PR #2548, opened 2026-02-20
chore: add info about packages in README.md

splincode:


README.md
```diff
@@ -48,6 +48,15 @@ For more, check out the [Contributing Guide](CONTRIBUTING.md).
 Maskito is a part of [Taiga UI](https://github.com/taiga-family/taiga-ui) libraries family which is backed and used by a
 large enterprise. This means you can rely on timely support and continuous development.
 
+| **Package**                                                    | **Version**                                                       | **README**                                                                        | **Downloads**                                                                                     |
```
nsbarsukov: I am not sure that we really need the such table in root README.

Maybe it is enough to just add badge here (with downloads for main `@maskito/core`) ?

Image

splincode: image

But it's convenient to see all packages by their download popularity

Issue #2541, opened 2026-02-16
🐞 - `maskitoParseDate` does not handle multi-char separators


Abreuvoir:
### Which package(s) are the source of the bug?

@maskito/kit

### Playground Link

https://stackblitz.com/edit/a5xi2nvw-4suxgjns?devToolsHeight=33&file=index.ts

### Description

Related to #2531 

`maskitoParseDate `can return valid date objects instead of null in some cases. 
It seems to be caused by the usage of a multi-character separator with dates under 4 digits; those should not allow strings not respecting the `YYYY` format.

### Maskito version

5.1.1

### Which browsers have you used?

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
Expected result

```ts
maskitoParseDate('16-01-20', { mode: 'dd/mm/yyyy', separator: '-' }) // null
maskitoParseDate('16--01--20', { mode: 'dd/mm/yyyy', separator: '--' }) // null

maskitoParseDate('16-01-2026', { mode: 'dd/mm/yyyy', separator: '-' }) // Date
maskitoParseDate('16--01--2026', { mode: 'dd/mm/yyyy', separator: '--' }) // Date
```

PR #2561, opened 2026-03-06
fix(kit): `maskitoParseDate` should return `null` for invalid `Date` string

### LINKED ISSUE #2558: 🐞 -  `maskitoParseDate` returns a valid Date object even when the input is not a date (invalid format not detected)

@maskito/kit

### Playground Link

https://stackblitz.com/edit/f3ljb2qe

### Description

When calling `maskitoParseDate()` with a string that is not a date at all, Maskito still returns a constructed JavaScript Date object, instead of returning null, throwing an error, or indicating invalid parsing.
Example:
```ts
maskitoParseDate("this-is-not-a-date", params)// → returns Date(0000-12-31T23:25:52.000Z)
```

Maybe I miss something and this is intentional ? I that case any way to detect an invalid format ?

### Maskito version

5.1.0

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
Fixes #2558
Fixes https://github.com/taiga-family/maskito/issues/2541

### DIFF
--- a/projects/kit/src/lib/masks/date/utils/parse-date.ts
+++ b/projects/kit/src/lib/masks/date/utils/parse-date.ts
@@ -6,7 +6,10 @@ export function maskitoParseDate(
     value: string,
     {mode, min = DEFAULT_MIN_DATE, max = DEFAULT_MAX_DATE}: MaskitoDateParams,
 ): Date | null {
-    if (value.length < mode.length) {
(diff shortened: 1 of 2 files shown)


zecka:
Awesome, thank you !

PR #2600, opened 2026-03-26
fix(kit): `maskitoParseNumber` incorrectly parses postfix containing digits

nsbarsukov:
Fixes 
* https://github.com/taiga-family/taiga-ui/issues/13612

### DIFF
--- a/projects/demo/src/pages/kit/number/helpers/parse-number-invalid-usage.md
+++ b/projects/demo/src/pages/kit/number/helpers/parse-number-invalid-usage.md
@@ -1,0 +1,5 @@
+```ts
+maskitoParseNumber('-42'); // -42 ✅
+maskitoParseNumber('> -42'); // 42 ❌
+maskitoParseNumber('> -42', {prefix: '> '}); // -42 ✅
+```
--- a/projects/demo/src/pages/kit/number/number-mask-doc.component.ts
+++ b/projects/demo/src/pages/kit/number/number-mask-doc.component.ts
@@ -61,6 +61,9 @@ export default class NumberMaskDocComponent implements GeneratorParams {
     protected readonly parseNumberAsBigIntTypeDemo =
         import('./helpers/parse-number-as-bigint-type.md');
 
+    protected readonly parseNumberInvalidUsageDemo =
+        import('./helpers/parse-number-invalid-usage.md');
+
     protected readonly stringifyNumberDemo = import('./helpers/stringify-number.md');
 
     protected readonly highPrecisionExample1: Record<string, TuiRawLoaderContent> = {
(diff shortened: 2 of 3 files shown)


projects/kit/src/lib/masks/number/utils/with-number-defaults.ts
```diff
@@ -0,0 +1,55 @@
+import {
+    CHAR_MINUS,
+    CHAR_NO_BREAK_SPACE,
+    CHAR_ZERO_WIDTH_SPACE,
+    DEFAULT_PSEUDO_MINUSES,
+} from '../../../constants';
+import {type MaskitoNumberParams} from '../number-params';
+import {validateDecimalPseudoSeparators} from './validate-decimal-pseudo-separators';
+
+export function withNumberDefaults({
```
splincode, line 10: What do you think

`normalizeNumberParams` or `resolveNumberParams`?

projects/kit/src/lib/masks/number/number-mask.ts
```diff
@@ -32,70 +25,28 @@ import {
...
-        prefix:
-            prefix.endsWith(decimalSeparator) && maximumFractionDigits > 0
-                ? `${prefix}${CHAR_ZERO_WIDTH_SPACE}`
-                : prefix,
-        minimumFractionDigits: Math.min(minimumFractionDigits, maximumFractionDigits),
-    };
+export function maskitoNumberOptionsGenerator(
+    optionalParams?: MaskitoNumberParams,
+): Required<MaskitoOptions> {
+    const params = withNumberDefaults(optionalParams);
```
splincode, line 33: why not 

```ts
const {minusSign, minusPseudoSigns, ...} = withNumberDefaults(optionalParams);
```

projects/kit/src/lib/masks/number/utils/with-number-defaults.ts
nsbarsukov: `withNumberDefaults` sounds good, does not it ?
Short and self-descriptive enough.

For example, Vue uses the same naming pattern
https://vuejs.org/guide/typescript/composition-api.html#props-default-values

projects/kit/src/lib/masks/number/number-mask.ts
nsbarsukov: But i should pass `params` in every processor below too

splincode: @nsbarsukov you can

```ts
const params = withNumberDefaults(optionalParams),
    {minusSign, minusPseudoSigns, ...} = params;
```

PR #2649, opened 2026-04-16
chore: add explicit untracked reads to effect guard

splincode:
https://angular.dev/guide/signals#reactive-context-and-async-operations

image


projects/angular/src/lib/tests/maskito.directive.spec.ts
nsbarsukov, line 1: Should we move the tests to `projects/angular/src/lib/tests`? This file and the existing `maskito.spec.ts` would go there — now that there's more than one, it makes sense to consolidate them.

nsbarsukov (approved):

PR #2663, opened 2026-04-22
chore: add `[alt]` for `img`

splincode:
image


projects/demo/src/app/modules/example-primary-tabs-icons/angular-logo.component.ts
```diff
@@ -3,7 +3,7 @@ import {PolymorpheusComponent} from '@taiga-ui/polymorpheus';
 
 @Component({
     selector: 'angular-logo',
-    template: '<img src="assets/icons/angular.svg" />',
+    template: '<img src="assets/icons/angular.svg" alt="logo" />',
```
nsbarsukov: Should it be `Angular logo`, `React logo` and etc. ?

nsbarsukov (approved):

PR #2711, opened 2026-05-18
chore(demo): update Taiga UI to `v5` 

nsbarsukov:


waterplea:
A few sizes are off when comparing to Taiga UI + maybe we can add icons to the navigation?

image
image

waterplea (approved):

PR #2718, opened 2026-05-19
chore(demo): use `tui-textfield[iconEnd]` instead of explicit import of `TuiIcon`

nsbarsukov:


projects/demo/src/pages/documentation/real-world-form/index.html
```diff
@@ -106,7 +106,10 @@ <h3 class="tui-form__header tui-form__header_margin-top_none">Real World Form</h
     </div>
 
     <div class="tui-form__row tui-form__row_multi-fields">
```
waterplea, line 108: Would be great to migrate from classes to `tuiForm` component as well as those global styles are no longer supported.

waterplea (approved):

PR #2765, opened 2026-06-09
chore(demo): use `.md` (instead of `.tsx`) for documentation example about `react-hook-form`

splincode:


projects/demo/package.json
```diff
@@ -31,6 +31,7 @@
         "browser-sync": "3.0.4",
         "react": "19.2.6",
         "react-dom": "19.2.6",
+        "react-hook-form": "7.78.0",
```
nsbarsukov: Overkill for only one single documentation tab ?

splincode: Why? This package is part of the documentation, so why can't we just build the project with it?

nsbarsukov: > This package is part of the documentation

This package is only part of the single not-executable file (not even actually used)

splincode: fixed

projects/demo/src/pages/frameworks/react/examples/3-react-hook-form/with-maskito-register.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-redundant-type-constituents -- because ts-ignore */
 import type {RefCallback} from 'react';
 // @ts-ignore
```
nsbarsukov: The single example tab contains too many noisy code comments that exist solely to satisfy our internal linters and **aren't relevant to users**.

Let's extend `TUI_DOC_EXAMPLE_CONTENT_PROCESSOR` with logic to strip out `eslint-disable` and `@ts-ignore [SKIP]` comments from the documentation tabs.

Image

nsbarsukov (approved):

PR #2775, opened 2026-06-11
chore(demo): restore previously removed dark mode button

nsbarsukov:


splincode (approved):

splincode:
image

It's as if the theme icon is less saturated than the others when hovered over.

PR #2873, opened 2026-09-04
fix(kit): date related masks pad incomplete date segments on separator insertion

### LINKED ISSUE #2132: 🚀 - Date related masks support zero-padding for incomplete segment on separator insertion

@maskito/kit

### Playground Link

https://maskito.dev/kit/date

### Description

Hello!

I didn't provide a stackblitz because we can easily reproduce it inside the doc (https://maskito.dev/kit/date)

I think the mask isn't correctly parsed, I tried to provide a fix but after 2h, I gave up 😄 

## Current Behavior

When we write 2000/6, it is directly parsed in 2000/06 and we can still write the rest of the mask. Fine !
When we write 2000/1, it is still correct but once I add my separator, nothing happens.

## Desired Behavior

I would like to be able to go to the next segment of the date when I manually insert the separator in the month or day segment. For example, writing `yyyy/m/` should render `yyyy/mm/` (zero-padded).
This only occurs for month with this mask. But with the following mask dd/mm/yyyy, I'm unable to write 1/1/2000 and get 01/01/2000. Instead, I have 11/1 🤔 

Some examples:

- `yyyy/m/` -> `yyyy/mm/`
- `d/m/yyyy` -> `dd/mm/yyyy`
- `d/m` -> `dd/mm`

Thank you in advance ! 🤗 

### Maskito version

3.8.0

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [ ] Windows
- [x] Linux
- [ ] iOS
- [ ] Android

splincode:
Fixes #2132


projects/demo-integrations/src/tests/kit/date-range/date-separator-zero-padding.cy.ts
```diff
@@ -0,0 +1,26 @@
...
+        cy.get('#demo-content input')
+            .should('be.visible')
+            .first()
+            .focus()
+            .type('01012000')
+            .should('have.value', FIRST_DATE)
+            .as('input');
+    });
+
+    it('Type 1.1. for the second date => 01.01.|', () => {
```
nsbarsukov, line 35: Let's add the same test for day-segment

```
type 1. for the second date => 01.|
```

___

Also add test for twice typed separator in a row

```
type 1 with subsequent two dots => 01.|
```

nsbarsukov: The same comments for `**/date-time/date-separator-zero-padding.cy.ts` and `*date/date-separator-zero-padding.cy.ts` specs

projects/kit/src/lib/utils/date/pad-incomplete-date-segment.ts
```diff
@@ -0,0 +1,20 @@
...
+}): string {
+    const dateSegments = dateString.split(dateSeparator);
+    const segmentIndex = dateSegments.length - 1;
+    const segment = dateSegments[segmentIndex] ?? '';
+    const templateSegment = dateModeTemplate.includes('d')
+        ? dateModeTemplate.split(dateSeparator)[segmentIndex]
+        : undefined;
+
+    return templateSegment && !templateSegment.includes('y') && /^[1-9]$/.test(segment)
+        ? `${dateString.slice(0, -segment.length)}${segment.padStart(templateSegment.length, '0')}`
```
nsbarsukov, line 16: Could/Should we use `parseDateString` & `toDateString` here ?

projects/kit/src/lib/utils/date/pad-incomplete-date-segment.ts
```diff
@@ -0,0 +1,20 @@
...
+    dateString: string;
+    dateModeTemplate: string;
+    dateSeparator: string;
+}): string {
+    const dateSegments = dateString.split(dateSeparator);
+    const segmentIndex = dateSegments.length - 1;
+    const segment = dateSegments[segmentIndex] ?? '';
+    const templateSegment = dateModeTemplate.includes('d')
+        ? dateModeTemplate.split(dateSeparator)[segmentIndex]
+        : undefined;
```
nsbarsukov: `mm/yy`, `mm/yyyy` modes ?

Add cypress tests for them too

projects/demo-integrations/src/tests/kit/date-range/date-separator-zero-padding.cy.ts
splincode: Done. Added cases for the day segment and for typing the separator twice in a row.

projects/kit/src/lib/utils/date/pad-incomplete-date-segment.ts
splincode: Done, added the same cases for Date and DateTime as well.

projects/kit/src/lib/utils/date/pad-incomplete-date-segment.ts
splincode: Added tests for `mm/yy` and `mm/yyyy` and fixed the helper to support modes without a day segment.

Regarding `parseDateString`, `toDateString`: I don't think they fit this case because `parseDateString` strips separators before parsing. For an incomplete value like `1.1.` it loses the segment boundaries, while here we need to know exactly which incomplete segment the user is editing.

nsbarsukov (approved):

